PR #207, opened 2024-08-16
Refactor statemachine

sdojjy:


scheduler/scheduler_balance.go
```diff
@@ -79,7 +78,6 @@ func (b *balanceScheduler) Schedule(
 		b.random, aliveCaptures, stateMachines, batchSize)
 	b.forceBalance = len(tasks) != 0
 	log.Info("balance scheduler generate tasks",
-		zap.String("id", b.id.String()),
```
CharlesCheung96, line 82: This field is used to distinguish between different schedulers.

sdojjy: I will add it back

PR #309, opened 2024-09-19
eventBroker: use dynamic stream to send watermark

asddongmen:


pkg/eventservice/event_broker.go
```diff
@@ -228,14 +238,56 @@ func (c *eventBroker) sendDDL(ctx context.Context, remoteID node.ID, e common.DD
...
+// 1. send the watermark to the dispatcher
+// 2. push the task to the task pool
+func (c *eventBroker) checkNeedScan(ctx context.Context, task scanTask) (bool, *common.DataRange, []common.DDLEvent) {
 	dataRange, needScan := task.dispatcherStat.getDataRange()
 	if !needScan {
-		return
+		return false, dataRange, nil
+	}
+	ddlEvents, endTs, err := c.schemaStore.FetchTableDDLEvents(dataRange.Span.TableID, dataRange.StartTs, dataRange.EndTs)
+	if err != nil {
```
hongyunyan, line 256: 这里要记得 return 吧？

pkg/eventservice/event_broker.go
```diff
@@ -88,6 +94,15 @@ func newEventBroker(
 	ctx, cancel := context.WithCancel(ctx)
 	wg := &sync.WaitGroup{}
 
+	option := dynstream.NewOption()
+	option.MaxPendingLength = 1
+	option.DropPolicy = dynstream.DropEarly
+	option.StreamCount = 512
+	option.SchedulerInterval = 500 * time.Millisecond
```
flowbehappy: Should not need to change the default setting

pkg/eventservice/event_broker.go
```diff
@@ -130,10 +144,18 @@ func (c *eventBroker) sendWatermark(
...
+	for i := 0; i < resolvedEventRetryTime; i++ {
+		select {
+		case c.messageCh <- resolvedEvent:
+			if counter != nil {
+				counter.Inc()
+			}
+			return
+		default:
+			time.Sleep(time.Millisecond * 10)
+		}
```
flowbehappy, line 156:
```suggestion
		select {
		case c.messageCh <- resolvedEvent:
			if counter != nil {
				counter.Inc()
			}
			return
		default:
		       // Drop resolved events if downstream cannot catch up with
		}
```

pkg/eventservice/event_broker.go
```diff
@@ -642,35 +688,50 @@ func (s *spanSubscription) onNewEvent(raw *common.RawKVEntry) {
...
 	// pendingTaskQueue is used to store the tasks that are waiting to be handled by the scan workers.
 	// The length of the pendingTaskQueue is equal to the number of the scan workers.
-	pendingTaskQueue []chan *scanTask
+	pendingTaskQueue []chan scanTask
 }
 
 func newScanTaskPool() *scanTaskQueue {
 	res := &scanTaskQueue{
-		pendingTaskQueue: make([]chan *scanTask, defaultScanWorkerCount),
+		pendingTaskQueue: make([]chan scanTask, defaultScanWorkerCount),
```
flowbehappy, line 713: I think we don't need to do partition any more

PR #520, opened 2024-11-13
use heap in dispatcher

CharlesCheung96:


CharlesCheung96:
do not use single heap for 100w itmes

PR #526, opened 2024-11-13
[WIP] eventService: reduce resolvedTs lag

asddongmen:


asddongmen:
image
Lag when replicating 100,000 tables

PR #678, opened 2024-12-13
eventCollertor: handle event concurrently

asddongmen:


downstreamadapter/eventcollector/event_collector.go
```diff
@@ -38,6 +39,11 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	receiveChanNum  = 32
```
flowbehappy, line 43: I think we should use the same var between event_collector and event_broker. So that they are always the same. And set this value to a large enough, e.g. `128`, to make sure we never have to change it between different versions.

asddongmen: Got.

PR #743, opened 2024-12-31
maintainer: fix remove changefeed and some data race

asddongmen:


asddongmen:
After investigation, it was found that the failure of pause + resume CI is related to the processing logic of DDL. 
Currently, the DDL processing logic is in the testing and bug-fixing phase. 
This PR primarily addresses some data race issues and optimizes the logic to avoid resending excessive unnecessary messages when removing changefeed. 
Therefore, it has been decided to merge this PR first, while DDL-related issues will be handled separately in the future.
cc @hongyunyan

PR #1228, opened 2025-04-15
codec: fix the open protocol flag

### LINKED ISSUE #1227: kafka integration test since open protocol flag is incorrect
Run the following SQL:
```
create table t(id int primary key)
insert into t values (1)
```

The tiflow result:
```
{"u":{"id":{"t":3,"h":true,"f":11,"v":1}}}
```

The ticdc result:
```
{"u":{"id":{"t":3,"h":true,"f":4099,"v":1}}}
```

3AceShowHand:

Issue Number: close #1227 

* Changing the column Flag type from uint to uint64 and updating related flag-check calls to use new helper functions (e.g. isBinary, isUnsigned, isPrimary).
* Introducing a column flags caching mechanism in the batch encoder to improve flag calculation efficiency.
* Adjusting encoding and decoding logic to properly pass the computed column flags.

#### Tests 

- Unit test
- Integration test

### Release note 

```release-note
`None`
```

### DIFF
--- a/pkg/sink/codec/open/decoder.go
+++ b/pkg/sink/codec/open/decoder.go
@@ -251,7 +251,7 @@ func buildColumns(
 			Value: value,
 		}
 		if _, ok := handleKeyColumns[name]; ok {
-			col.Flag = mysql.PriKeyFlag
+			col.Flag |= binaryFlag
 		}
 		result[name] = col
 	}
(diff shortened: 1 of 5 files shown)


pkg/sink/codec/open/encoder.go
```diff
@@ -61,13 +64,23 @@ func (d *batchEncoder) Clean() {
 	}
 }
 
+func (d *batchEncoder) fetchColumnFlags(e *commonEvent.RowEvent) map[string]uint64 {
+	result, ok := d.columnFlagsCache[e.GetTableID()]
```
wk989898, line 68: Why cache column flag? It looks better to use a map to convert flag into columnFlag.

pkg/sink/codec/open/decoder.go
```diff
@@ -251,7 +251,7 @@ func buildColumns(
 			Value: value,
 		}
 		if _, ok := handleKeyColumns[name]; ok {
-			col.Flag = mysql.PriKeyFlag
+			col.Flag |= binaryFlag
```
wk989898, line 254: Why use a binary flag?

3AceShowHand: it's found a binary flag.

pkg/sink/codec/open/encoder.go
3AceShowHand: the columnFlag is calculated by the flag, and cache it to prevent calculation on each message

pkg/sink/codec/open/message.go
```diff
@@ -206,3 +207,117 @@ func (m *messageRow) decode(data []byte) {
...
+		for _, idxCol := range idxInfo.Columns {
+			flag := result[idxCol.Name.O]
+			if idxInfo.Primary {
+				flag |= primaryKeyFlag
+			} else if idxInfo.Unique {
+				flag |= uniqueKeyFlag
+			}
+			if len(idxInfo.Columns) > 1 {
+				flag |= multipleKeyFlag
+			}
```
wk989898, line 314: Please keep same with simple protocol. https://github.com/pingcap/ticdc/blob/39a11cc09049b6685fe647a7ba603728150b10f6/pkg/sink/codec/simple/decoder.go#L413-L426

wk989898 (approved):

PR #1240, opened 2025-04-17
http: fix the api capture list output does not match the old architecture

### LINKED ISSUE #1241: http open api capture list output does not match the previous implementation.
Image

3AceShowHand:

Issue Number: close #1241 

#### Tests 

- Unit test
- Integration test

### Release note 

```release-note
`None`
```

### DIFF
--- a/api/v2/model.go
+++ b/api/v2/model.go
@@ -1184,7 +1184,7 @@ type ServerStatus struct {
 type Capture struct {
 	ID string `json:"id"`
 	// IsCoordinator is true if the capture is the coordinator of the TiCDC cluster
-	// We make its json key as `is_owner` to keep the compatibility with old TiCDC.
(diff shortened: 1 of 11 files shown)


asddongmen (approved):

api/v2/model.go
```diff
@@ -1185,9 +1185,9 @@ type Capture struct {
 	ID string `json:"id"`
 	// IsCoordinator is true if the capture is the coordinator of the TiCDC cluster
 	// We make its json key as `is_owner` to keep the compatibility with old TiCDC.
```
asddongmen, line 1187: Please also change this comment.

3AceShowHand, line 1187:
```suggestion
	// We make its json key as `is-owner` to keep the compatibility with old TiCDC.
```

Issue #1061, opened 2025-03-04
Add redo support


hicqu:
### Is your feature request related to a problem?

Redo should be introduced into TiCDC.

### Describe the feature you'd like
* [x] Add redo package #1084
* [x] Add redo dispatcher #1480 #1567
* [ ] Refactor redo apply #2167
* [x] Add redo sink #1461
* [x] Rewrite redo writer #1442
* [ ] Improve scheduler performance https://github.com/pingcap/ticdc/pull/1652
* [x] Add metrics about redo https://github.com/pingcap/ticdc/pull/2028

wk989898:
Since we flush the redo meta with the background thread, some events for which the commit-ts is greater than the redo meta resolved-ts may be flushed downstream. These events are duplicated.
https://github.com/pingcap/ticdc/blob/3c2d89aa10b249b1bad39345d943109032fb7b69/downstreamadapter/dispatchermanager/dispatcher_manager_redo.go#L249-L255
https://github.com/pingcap/ticdc/blob/3c2d89aa10b249b1bad39345d943109032fb7b69/downstreamadapter/sink/redo/meta.go#L346-L370

PR #1344, opened 2025-05-21
*: move spanz to common package

wk989898:

Issue Number: ref #1061

#### Tests 

- Unit test
- Integration test

### Release note 

```release-note`None`.
```


downstreamadapter/dispatcher/dispatcher.go
```diff
@@ -451,9 +452,8 @@ func (d *Dispatcher) PassBlockEventToSink(event commonEvent.BlockEvent) {
 }
 
 func isCompleteSpan(tableSpan *heartbeatpb.TableSpan) bool {
```
hongyunyan, line 454: can we use common.IsCompleteSpan directly

downstreamadapter/downstream_performance_test.go
```diff
@@ -53,133 +42,133 @@ func initContext(serverId node.ID) {
...
-				finishCount++
-				if finishCount == databaseCount {
-					log.Warn("All data consuming is finished")
-					return
-				}
-			}
-		}
-	}
-}
+// func pushDataIntoDispatchers(dispatcherIDSet map[common.DispatcherID]interface{}, helper *commonEvent.EventTestHelper) {
```
hongyunyan, line 45: This change is no need in this pr?

downstreamadapter/eventcollector/event_collector_test.go
```diff
@@ -85,6 +81,60 @@ func (m *mockEventDispatcher) HandleEvents(dispatcherEvents []dispatcher.Dispatc
 	return false
 }
 
+func (m *mockEventDispatcher) HandleCheckpointTs(checkpointTs uint64) {
```
hongyunyan, line 84: do we need this change here?

downstreamadapter/dispatcher/dispatcher.go
wk989898: In fact, there are two `IsCompleteSpans` in dispatcher and coordinator. However, the implementation of these two functions is different.

downstreamadapter/downstream_performance_test.go
wk989898: The unit test always fails, so I commented out the code.

hongyunyan (approved):

lidezhu (approved):

PR #1588, opened 2025-07-28
eventcollector, eventservice: add congestion control mechanism

### LINKED ISSUE #1404: CDC node oom when has 4 changefeeds replicating 100 same tables
Image

Image

Image

* Pause all 4 changefeeds, and insert 20,000,000 rows on the upstream database.
* Resume all changefeeds, not changefeeds write data to the target.
* The CPU usage is full, and memory takes more than 29GB. There is on cdc instances, 16c, 32g.

Image

Accoriding to the profiling, most of the memory cost comes from the mounter decode events.

3AceShowHand:

Issue Number: close #1404 

* Introduce congestion control mechanism, event collector report available memory to the event broker periodically.
* Use decoded dml event size to calculate the scan limit, instead of the raw entry, this is more accurate.
* update grafana to record more metrics about the scanner.
* When scanning data, break the scanning if exceed the limit. Only output resolved-ts if the next row has different commit-ts, to keep compatible to the async-commit transactions.
* When build the iterator, also consider the last scanned start-ts, to prevent rescanning the same data.
* Kafka sink send heartbeat periodically to prevent broken pipe.

#### Tests 

- Unit test
- Integration test

### Release note 

```release-note
`None`
```


downstreamadapter/eventcollector/event_collector.go
```diff
@@ -415,6 +419,81 @@ func (c *EventCollector) runDispatchMessage(ctx context.Context, inCh <-chan *me
...
+				msg := messaging.NewSingleTargetMessage(serverID, messaging.EventServiceTopic, m)
+				if err := c.mc.SendCommand(msg); err != nil {
+					log.Warn("send congestion control message failed", zap.Error(err))
+				}
+			}
+		}
+	}
+}
+
+func (c *EventCollector) newCongestionControlMessages() map[node.ID]*event.CongestionControl {
```
flowbehappy, line 442: Please explain the basic idea of the algorithm

pkg/eventservice/event_broker.go
```diff
@@ -570,6 +584,18 @@ func (c *eventBroker) doScan(ctx context.Context, task scanTask) {
 	metricEventBrokerScanTaskCount.Inc()
 }
 
+func allocQuota(quota *atomic.Uint64, nBytes uint64) {
+	for {
+		available := quota.Load()
+		if available < nBytes {
+			continue
```
flowbehappy, line 591: So the thread is looped here until available quota? I think should avoid that.

pkg/eventservice/event_broker.go
```diff
@@ -497,28 +507,32 @@ func (c *eventBroker) doScan(ctx context.Context, task scanTask) {
...
+			zap.String("changefeed", changefeedID.String()), zap.String("remote", remoteID.String()))
+	}
+	available := item.(*atomic.Uint64)
+
+	if available.Load() <= memoryQuotaLowThreshold {
+		task.resetScanLimit()
+	}
+
+	sl := c.calculateScanLimit(task)
+	allocQuota(available, uint64(sl.maxDMLBytes))
```
flowbehappy, line 528: I think the scan thread should not wait here until available quota.

downstreamadapter/eventcollector/event_collector.go
```diff
@@ -415,6 +419,86 @@ func (c *EventCollector) runDispatchMessage(ctx context.Context, inCh <-chan *me
...
+		return nil
+	}
+
+	// calculate each changefeed's available memory quota for each node
+	// by the proportion of the dispatcher on each node.
+	// this is not accurate, we should also consider each node's workload distribution.
+	proportions := make(map[common.ChangeFeedID]map[node.ID]uint64)
+	c.dispatcherMap.Range(func(k, v interface{}) bool {
+		stat := v.(*dispatcherStat)
+		eventServiceID := stat.connState.getEventServiceID()
```
lidezhu, line 462: eventServiceID may be empty when the dispatcher is initializing.

downstreamadapter/eventcollector/event_collector.go
```diff
@@ -415,6 +419,86 @@ func (c *EventCollector) runDispatchMessage(ctx context.Context, inCh <-chan *me
...
+		changefeedID := stat.target.GetChangefeedID()
+
+		holder, ok := proportions[changefeedID]
+		if !ok {
+			holder = make(map[node.ID]uint64, len(availables))
```
lidezhu, line 467: len(availables)) seems incorrect. When dealing with a large number of changefeeds, this approach could lead to large memory wast.

logservice/eventstore/event_store.go
```diff
@@ -909,10 +909,12 @@ func (iter *eventStoreIter) Next() (*common.RawKVEntry, bool) {
 		iter.innerIter.Next()
 	}
 	isNewTxn := false
-	if iter.prevCommitTs == 0 || (rawKV.StartTs != iter.prevStartTs || rawKV.CRTs != iter.prevCommitTs) {
+	// 2 PC transactions have different startTs and commitTs.
+	// async-commit transactions have different startTs and may have the same commitTs.
+	// so use the startTs to determine whether it is a new transaction.
+	if iter.prevStartTs == 0 || rawKV.StartTs != iter.prevStartTs {
```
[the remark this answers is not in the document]
3AceShowHand: different txn have different start-ts, but may have same commit-ts.

Previous implementation check on using the commit-ts, treat multiple different txns as the same one. It leads to hard hit limit checker. During test we found that, the limit is set to 4MB, but more than 13MB data is scanned, since those txns have the same commit-ts, different start-ts.

downstreamadapter/dispatcher/dispatcher.go
```diff
@@ -254,7 +254,6 @@ func (d *Dispatcher) HandleEvents(dispatcherEvents []DispatcherEvent, wakeCallba
 	for _, dispatcherEvent := range dispatcherEvents {
 		log.Debug("dispatcher receive all event",
 			zap.Stringer("dispatcher", d.id),
-			zap.String("eventType", commonEvent.TypeToString(dispatcherEvent.Event.GetType())),
 			zap.Any("event", dispatcherEvent.Event))
```
3AceShowHand, line 257: event type can be know from the event field, so no need to print this, duplicate field

asddongmen: Please restore this change. The eventType printed in zap.Any("event", dispatcherEvent.Event) is of int type and is not human - readable.

3AceShowHand: I found that it's human-readable, since the `event` first part is something like `ResolvedEvent`.

logservice/eventstore/event_store.go
```diff
@@ -687,16 +686,19 @@ func (e *eventStore) GetIterator(dispatcherID common.DispatcherID, dataRange com
 	}
 
 	// convert range before pass it to pebble: (startTs, endTs] is equal to [startTs + 1, endTs + 1)
-	start := EncodeKeyPrefix(uint64(subStat.subID), stat.tableSpan.TableID, dataRange.StartTs+1)
+	var start []byte
```
3AceShowHand, line 689: This is a critical change

downstreamadapter/eventcollector/event_collector.go
lidezhu: `len(availables))` is the number of changefeeds. But I think what you need is the number of cdc nodes.

downstreamadapter/eventcollector/event_collector.go
```diff
@@ -512,6 +516,87 @@ func (c *EventCollector) runDispatchMessage(ctx context.Context, inCh <-chan *me
...
+	for changefeedID, total := range availables {
+		proportion := proportions[changefeedID]
+		var sum uint64
+		for _, portion := range proportion {
+			sum += portion
+		}
+
+		for nodeID, portion := range proportion {
+			ratio := float64(portion) / float64(sum)
+			quota := uint64(float64(total) * ratio)
```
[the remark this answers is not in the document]
3AceShowHand: This sounds good to me.

asddongmen (approved):

flowbehappy (approved):

PR #1618, opened 2025-08-07
fix(DMLEvent): prevent DMLEvent inconsistency on zero-row decodes and refactor helpers

### LINKED ISSUE #1617: DMLEvent state becomes inconsistent when DecodeToChunk returns zero rows
#### Summary

A bug exists in `pkg/common/event/dml_event.go` where the `DMLEvent.AppendRow` method incorrectly modifies the event's state when its underlying row-decoding function (`DecodeToChunk`) returns a row count of zero.

#### Problem Description

Currently, `DMLEvent.AppendRow` increments the event's `Length` and `ApproximateSize` properties regardless of the number of rows actually decoded and returned by `DecodeToChunk`.

While `DecodeToChunk` is expected to return a positive row count in normal operation, it can return `0` in exceptional cases or, as seen in the test suite, in mocked environments. When `count` is `0`, the following inconsistency occurs:
* `t.Length` is incremented by 1.
* No corresponding row data is appended to the `t.RowTypes` and `t.RowKeys` slices.

This creates a corrupted `DMLEvent` where its `Length` indicates the presence of a row, but the data slices are empty. Downstream consumers of this event are likely to encounter `index out of range` panics or process data incorrectly.

#### Expected Behavior

The properties of a `DMLEvent` (`Length`, `ApproximateSize`, `RowTypes`, `RowKeys`, `Checksum`) should only be modified if the number of decoded rows (`count`) is greater than zero.

#### Actual Behavior

`DMLEvent.Length` and `ApproximateSize` are unconditionally incremented after every call to `AppendRow`, even when the decoded row count is zero, leading to an inconsistent internal state.

wlwilliamx:

Issue Number: close #1617

This PR addresses a bug that causes `DMLEvent` to enter an inconsistent state. It also introduces several code quality improvements for better readability and maintenance.

#### What this PR does and why it's needed

The core issue is that `DMLEvent.AppendRow` would increment its `Length` counter even if the underlying call to `DecodeToChunk` returned zero rows. This is an abnormal but possible scenario that leads to a corrupted event object, risking panics downstream.

This PR makes the system more robust by handling this exceptional case correctly. It also refactors related code for better clarity.

#### Detailed Changes

* **Robust Handling in `DMLEvent.AppendRow`**
    * The logic for appending row data and incrementing event counters (`Length`, `ApproximateSize`) is now enclosed within an `if count > 0` block. This is the primary fix that ensures the event's state is only updated when new rows are actually present.
    * An `else` block has been added to handle the abnormal `count == 0` case. It logs a warning to make such occurrences visible for debugging, noting that this is primarily expected in test environments.

* **Code Quality Refactoring in `RawKVEntry`**
    * New helper methods, `IsDelete()` and `IsInsert()`, have been added to `RawKVEntry`. These methods encapsulate the logic for identifying an entry's operation type, improving readability and centralizing the logic.
    * The `if raw.OpType == common.OpTypeDelete` check in `DMLEvent.AppendRow` has been updated to use the cleaner `raw.IsDelete()` method.

* **Correction of Test Mock**
    * In `pkg/eventservice/event_scanner_test.go`, the `mockMounter`'s implementation of `DecodeToChunk` has been updated.
    * Previously, it always returned a count of `0`, which constantly triggered the now-handled abnormal case.
    * It now returns realistic row counts (2 for an update, 1 for other DML types) to properly test the normal execution path and prevent unnecessary warnings during tests.

#### Tests 

- Unit test
- No code

No.

No.

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/common/event/dml_event.go
+++ b/pkg/common/event/dml_event.go
@@ -339,7 +339,7 @@ func (t *DMLEvent) AppendRow(raw *common.RawKVEntry,
 		tableInfo *common.TableInfo, chk *chunk.Chunk) (int, *integrity.Checksum, error),
 ) error {
 	rowType := RowTypeInsert
-	if raw.OpType == common.OpTypeDelete {
+	if raw.IsDelete() {
 		rowType = RowTypeDelete
 	}
 	if raw.IsUpdate() {
@@ -349,17 +349,22 @@ func (t *DMLEvent) AppendRow(raw *common.RawKVEntry,
 	if err != nil {
 		return err
 	}
-	for range count {
-		t.RowTypes = append(t.RowTypes, rowType)
+	if count > 0 {
+		for range count {
+			t.RowTypes = append(t.RowTypes, rowType)
 
-		keyCopy := make([]byte, len(raw.Key))
-		copy(keyCopy, raw.Key)
-		t.RowKeys = append(t.RowKeys, keyCopy)
-	}
-	t.Length += 1
-	t.ApproximateSize += raw.GetSize()
-	if checksum != nil {
-		t.Checksum = append(t.Checksum, checksum)
+			keyCopy := make([]byte, len(raw.Key))
+			copy(keyCopy, raw.Key)
+			t.RowKeys = append(t.RowKeys, keyCopy)
+		}
+		t.Length += 1
+		t.ApproximateSize += raw.GetSize()
+		if checksum != nil {
+			t.Checksum = append(t.Checksum, checksum)
+		}
+	} else {
+		// Only in test, the count will be 0.
+		log.Warn("DMLEvent.AppendRow: no rows decoded from the raw KV entry", zap.Any("raw", raw))
 	}
 	return nil
 }
--- a/pkg/eventservice/event_scanner_test.go
+++ b/pkg/eventservice/event_scanner_test.go
@@ -43,7 +43,11 @@ func makeDispatcherReady(disp *dispatcherStat) {
 }
 
 func (m *mockMounter) DecodeToChunk(rawKV *common.RawKVEntry, tableInfo *common.TableInfo, chk *chunk.Chunk) (int, *integrity.Checksum, error) {
-	return 0, nil, nil
+	if rawKV.IsUpdate() {
+		return 2, nil, nil
+	} else {
+		return 1, nil, nil
+	}
 }
 
 func TestEventScanner(t *testing.T) {
--- a/pkg/common/kv_entry.go
+++ b/pkg/common/kv_entry.go
@@ -63,9 +63,17 @@ func (v *RawKVEntry) IsResolved() bool {
 	return v.OpType == OpTypeResolved
 }
 
+func (v *RawKVEntry) IsDelete() bool {
+	return v.OpType == OpTypeDelete
+}
+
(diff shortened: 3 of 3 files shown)


pkg/common/event/dml_event.go
```diff
@@ -349,17 +349,22 @@ func (t *DMLEvent) AppendRow(raw *common.RawKVEntry,
...
+			t.RowKeys = append(t.RowKeys, keyCopy)
+		}
+		t.Length += 1
+		t.ApproximateSize += raw.GetSize()
+		if checksum != nil {
+			t.Checksum = append(t.Checksum, checksum)
+		}
+	} else {
+		// Only in test, the count will be 0.
+		log.Warn("DMLEvent.AppendRow: no rows decoded from the raw KV entry", zap.Any("raw", raw))
```
wk989898, line 367:
```suggestion
		}
		return nil
	}
	log.Panic("DMLEvent.AppendRow: no rows decoded from the raw KV entry", zap.Any("raw", raw))
```

pkg/common/event/dml_event.go
```diff
@@ -349,17 +349,21 @@ func (t *DMLEvent) AppendRow(raw *common.RawKVEntry,
...
+			copy(keyCopy, raw.Key)
+			t.RowKeys = append(t.RowKeys, keyCopy)
+		}
+		t.Length += 1
+		t.ApproximateSize += raw.GetSize()
+		if checksum != nil {
+			t.Checksum = append(t.Checksum, checksum)
+		}
+	} else {
+		log.Panic("DMLEvent.AppendRow: no rows decoded from the raw KV entry", zap.Any("raw", raw))
```
3AceShowHand, line 366: panic immediate if count == 0, remove the else

pkg/eventservice/event_scanner_test.go
```diff
@@ -43,7 +43,11 @@ func makeDispatcherReady(disp *dispatcherStat) {
 }
 
 func (m *mockMounter) DecodeToChunk(rawKV *common.RawKVEntry, tableInfo *common.TableInfo, chk *chunk.Chunk) (int, *integrity.Checksum, error) {
-	return 0, nil, nil
+	if rawKV.IsUpdate() {
+		return 2, nil, nil
+	} else {
```
3AceShowHand, line 48: remove the else, it's redundant

pkg/common/event/dml_event.go
wlwilliamx: done

pkg/eventservice/event_scanner_test.go
wlwilliamx: it's necessary for some unit test cases so I add this.

pkg/common/event/dml_event.go
```diff
@@ -338,8 +338,12 @@ func (t *DMLEvent) AppendRow(raw *common.RawKVEntry,
 		rawKv *common.RawKVEntry,
 		tableInfo *common.TableInfo, chk *chunk.Chunk) (int, *integrity.Checksum, error),
 ) error {
+	if raw.Value == nil && raw.OldValue == nil {
```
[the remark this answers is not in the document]
wlwilliamx: done

PR #2040, opened 2025-09-10
feat(*): Implement `force-replicate` and `ignore_ineligible_table` in the filter and refactor eligibility checks

### LINKED ISSUE #2081: Enhance Table Eligibility Logic and Properly Implement `force-replicate` and `ignore_ineligible_table`
Currently, the system has several limitations regarding the handling of tables without eligible primary or unique keys:

1.  **Inaccurate Eligibility Check:** The existing logic for determining if a table is "eligible" for replication is not comprehensive. It primarily checks for `PKIsHandle` but doesn't correctly handle tables with composite primary keys or unique keys on non-nullable columns. This leads to tables being incorrectly marked as ineligible.
2.  **Inefficient `force-replicate` Implementation:** The previous `force-replicate` logic was implemented at the sink level (specifically for MySQL). For every DML operation on a table without a primary key, it would iterate through all columns of the row to construct a `WHERE` clause. This is highly inefficient, especially for tables with many columns, and adds significant processing overhead to the sink.
3.  **Lack of Configurability:** There is no clear way to control the replication behavior for ineligible tables. Users cannot choose to forcefully replicate them or to simply ignore them and continue replicating other tables, which can lead to replication errors or unwanted behavior.

#### Proposed Solution

To address these issues, we propose a comprehensive refactoring of the table eligibility and replication logic:

1.  **Improve Eligibility Detection:**
    * Implement a robust function (`originalHasPKOrNotNullUK`) that correctly identifies a table's eligibility by checking for:
        * Tables where the primary key is the handle (`PKIsHandle`).
        * Tables with composite primary keys.
        * Tables with unique indexes where all indexed columns are defined as `NOT NULL`.
    * Cache this eligibility status in a new field (`HasPKOrNotNullUK`) within the `TableInfo` struct to avoid re-computation.

2.  **Centralize Eligibility Enforcement in the Filter:**
    * Move all eligibility checks from disparate parts of the codebase (like the sink and table selection) into the `filter` component.
    * The `filter` will now check the eligibility of a table at the very beginning of processing for both DML and DDL events.

3.  **Properly Implement `force-replicate` and `ignore_ineligible_table`:**
    * These configuration options will be respected by the `filter`.
    * `force_replicate: true`: Allows DML and DDL for tables without eligible keys to pass through the filter.
    * `ignore_ineligible_table: true`: Will cause the filter to silently discard DML and DDL events for ineligible tables.
    * This approach is more efficient as the decision to replicate is made early, preventing ineligible events from propagating through the system.

4.  **Remove Inefficient Sink Logic:**
    * Completely remove the old `force-replicate` parameter and its associated column-iterating logic from the MySQL sink's SQL builder. The sink's responsibility will be simplified to only constructing and executing SQL for events it receives.

5.  **Implement Verification API:**
    * Implement the previously dummy `VerifyTable` API endpoint to use this new logic, allowing users to check which tables in their configuration are considered eligible or ineligible before creating a changefeed.

wlwilliamx:

Issue Number: close #2081

This PR introduces a complete overhaul of how the system handles table eligibility and the `force-replicate` feature. The core idea is to centralize this logic within the `filter` for greater efficiency and correctness, while also introducing a new `ignore_ineligible_table` option.

The previous `force-replicate` implementation at the MySQL sink level was highly inefficient, as it required iterating over all columns for every DML operation to build a `WHERE` clause. This PR removes that logic entirely and replaces it with an upfront check in the `filter`.

#### Key Changes

* **Enhanced Eligibility Check (`pkg/common/table_info.go`)**
    * A new function, `originalHasPKOrNotNullUK`, has been added to accurately determine if a table has a valid primary key or a non-null unique key. This handles composite keys correctly.
    * The result is cached in a new boolean field, `HasPKOrNotNullUK`, on the `TableInfo` struct for efficient access.
    * The `IsEligible` method now uses this cached field.

* **Centralized Logic in Filter (`pkg/filter/filter.go`)**
    * The `filter` now performs an eligibility check at the beginning of `ShouldIgnoreDML` and `ShouldIgnoreDDL`.
    * If a table is not eligible, its events are discarded early, preventing them from being processed further. This respects the `forceReplicate` flag, which can override this behavior.
    * The old, redundant eligibility check in `ShouldIgnoreTable` has been removed.
    * The `ShouldIgnoreDDL` interface has been updated to accept a `tableInfo` object to perform this check.

* **Removal of Inefficient Sink Logic (`pkg/sink/mysql/`)**
    * The `forceReplicate` parameter has been removed from the MySQL sink's `Config`, `buildDelete`, `buildUpdate`, and `whereSlice` functions.
    * The logic in `whereSlice` that iterated over all table columns when a key was not found has been deleted. This resolves a major performance bottleneck.

* **API Implementation (`api/v2/changefeed.go`)**
    * The `VerifyTable` API endpoint is now fully implemented. It uses the new eligibility logic to report which tables are eligible and ineligible based on the provided changefeed configuration, including `forceReplicate` and `ignore_ineligible_table`.

* **Testing (`pkg/filter/filter_test.go`)**
    * Added a new `TestIsEligible` test suite to validate the new eligibility logic against various table structures (with PK, with UK, with nullable UK, views, etc.) and the `forceReplicate` setting.

#### Tests 

- Unit test
- Integration test
- Manual test

### Release note 

```release-note
None
```


api/v2/changefeed.go
```diff
@@ -319,9 +319,64 @@ func (h *OpenAPIV2) ListChangeFeeds(c *gin.Context) {
...
+		return
+	}
+	err = replicaCfg.ValidateAndAdjust(sinkURIParsed)
+	if err != nil {
+		_ = c.Error(errors.WrapError(errors.ErrInvalidReplicaConfig, err))
+		return
+	}
+
+	scheme := sinkURIParsed.Scheme
+	topic := strings.TrimFunc(sinkURIParsed.Path, func(r rune) bool {
```
asddongmen, line 347: There is a function named 
```
// GetTopic returns the topic name from the sink URI.
func GetTopic(sinkURI *url.URL) (string, error) {
	topic := strings.TrimFunc(sinkURI.Path, func(r rune) bool {
		return r == '/'
	})
	if topic == "" {
		return "", errors.ErrKafkaInvalidConfig.GenWithStack("no topic is specified in sink-uri")
	}
	return topic, nil
}
```

api/v2/changefeed.go
```diff
@@ -1270,6 +1325,11 @@ func getVerifiedTables(
 	if err != nil {
 		return nil, nil, err
 	}
+	log.Info("verifyTables completed",
```
asddongmen, line 1328: Remove this logs.

api/v2/changefeed.go
```diff
@@ -319,9 +319,64 @@ func (h *OpenAPIV2) ListChangeFeeds(c *gin.Context) {
...
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+	ineligibleTables, eligibleTables, err := getVerifiedTables(ctx, replicaCfg, kvStorage, cfg.StartTs, scheme, topic, protocol)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+	log.Info("verify table", zap.Strings("ineligibleTables", ineligibleTables), zap.Strings("eligibleTables", eligibleTables),
```
asddongmen, line 364: Please do not logs all the tables. It can be huge number of tables out there and flood the log file.

pkg/common/table_info.go
```diff
@@ -356,7 +360,56 @@ func (ti *TableInfo) IsEligible(forceReplicate bool) bool {
 	if ti.IsView() {
 		return true
 	}
-	return len(ti.columnSchema.HandleKeyIDs) != 0
+	return ti.HasPKOrNotNullUK
+}
+
+func originalHasPKOrNotNullUK(tableInfo *model.TableInfo) bool {
```
asddongmen, line 366: Please add an unit test to cover this method.

pkg/filter/filter.go
```diff
@@ -189,6 +195,14 @@ func (f *filter) Verify(tableInfos []*common.TableInfo) error {
 	// return f.dmlExprFilter.verify(tableInfos)
 }
 
+func (f *filter) isEligible(tableInfo *common.TableInfo) bool {
+	if tableInfo == nil {
+		// This should happen in test only
```
asddongmen, line 200:
```suggestion
		// This should only happens in test
```

pkg/filter/filter.go
```diff
@@ -113,9 +113,14 @@ func NewFilter(cfg *config.FilterConfig, tz string, caseSensitive bool, forceRep
 
 // ShouldIgnoreDML checks if a DML event should be ignore by conditions below:
 // 0. By startTs.
-// 1. By table name.
-// 2. By type.
+// 1. By eligibility of the table.
+// 2. By table name.
+// 3. By type.
 func (f *filter) ShouldIgnoreDML(dmlType common.RowType, preRow, row chunk.Row, tableInfo *common.TableInfo, startTs uint64) (bool, error) {
+	if !f.isEligible(tableInfo) {
```
asddongmen, line 120: This check is unnecessary. If `forceReplicate=false`, ineligible tables should not be scheduled for replication in the first place.
Performing the check here would cause confusion.

pkg/common/table_info.go
wlwilliamx: I added it in filter_test.go

asddongmen: You should add an unit test in this package to improve the test coverage rate.

pkg/filter/filter.go
wlwilliamx:
```go
func (f *filter) isEligible(tableInfo *common.TableInfo) bool {
	if tableInfo == nil {
		// This should happen in test only
		return true
	}
	return tableInfo.IsEligible(f.forceReplicate)
}
```

This is the implementation of the isEligible function. Here, the forceReplicate configuration is passed to TableInfo.IsEligible. Because the forceReplicate configuration is stored in the filter, the isEligible function needs to be called, and this is the earliest place where DML can be filtered.

api/v2/changefeed.go
wlwilliamx: done

api/v2/changefeed.go
wlwilliamx: done

pkg/common/table_info.go
wlwilliamx: It's already covered in TestIsEligible.

wlwilliamx:
pull-cdc-mysql-integration-light failed due to https://github.com/pingcap/ticdc/issues/2226

asddongmen (approved):

PR #2737, opened 2025-10-20
*: resolve deadlock when enable syncpoint

### LINKED ISSUE #2733: mysql changefeed with sync-point enabled may cause deadlock due to the memory quota full
Image

The memory quota cannot be released, so it's full, and cause cannot send more events.

asddongmen:

Issue Number: close #2733 

#### Tests 

- Unit test
- Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/eventservice/event_broker.go
+++ b/pkg/eventservice/event_broker.go
@@ -593,7 +593,6 @@ func (c *eventBroker) doScan(ctx context.Context, task scanTask) {
 		return
 	}
 	available := item.(*atomic.Uint64)
-
 	if available.Load() < c.scanLimitInBytes {
 		task.resetScanLimit()
 	}
--- a/utils/dynstream/interfaces.go
+++ b/utils/dynstream/interfaces.go
@@ -182,8 +182,7 @@ type DynamicStream[A Area, P Path, T Event, D Dest, H Handler[A, P, T, D]] inter
 	RemovePath(path P) error
 
 	// ReleasePath releases all the events of the path.
-	// After this call return, future events with the path will be dropped, including events which are already in the stream.
-	// If the path doesn't exist, it will return ErrorTypeNotExist.
+	// After this call return, events in the path will be dropped.
 	Release(path P)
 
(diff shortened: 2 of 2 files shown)


pkg/eventservice/event_broker.go
```diff
@@ -589,11 +601,18 @@ func (c *eventBroker) doScan(ctx context.Context, task scanTask) {
 	sl := c.calculateScanLimit(task)
 	ok = allocQuota(available, uint64(sl.maxDMLBytes))
 	if !ok {
-		log.Debug("not enough memory quota, skip scan",
+		log.Debug("changefeed available memory quota is not enough, skip scan",
 			zap.String("changefeed", changefeedID.String()),
 			zap.String("remote", remoteID.String()),
 			zap.Uint64("available", available.Load()),
 			zap.Uint64("required", uint64(sl.maxDMLBytes)))
+		c.sendSignalResolvedTs(task)
```
lidezhu: why need `sendSignalResolvedTs` here?

utils/dynstream/stream.go
```diff
@@ -316,28 +323,26 @@ type pathInfo[A Area, P Path, T Event, D Dest, H Handler[A, P, T, D]] struct {
 	// guaranteed to see the memory change of this field.
 	removed atomic.Bool
 	// The path is blocked by the handler.
-	blocking bool
+	blocking atomic.Bool
```
lidezhu: Why change it to an atomic variable? Atomic variable means more complex usage and is more error prone. So I want to know why it is changed in this pr.

utils/dynstream/interfaces.go
```diff
@@ -181,11 +181,16 @@ type DynamicStream[A Area, P Path, T Event, D Dest, H Handler[A, P, T, D]] inter
 	// If the path doesn't exist, it will return ErrorTypeNotExist.
 	RemovePath(path P) error
 
+	// ReleasePath releases all the events of the path.
+	// After this call return, future events with the path will be dropped, including events which are already in the stream.
```
lidezhu: future events is also dropped? and when the path can normally handle events?

PR #2752, opened 2025-10-21
server: avoid server hang when sub module close

wk989898:

Issue Number: ref #2750

This pull request implements a crucial enhancement to the CDC server's shutdown process by introducing a timeout mechanism. This change ensures that the server will not get stuck indefinitely if certain internal modules fail to terminate promptly, thereby improving the robustness and reliability of the server's lifecycle management.
#### Tests 

- Unit test
- Integration test

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cmd/cdc/server/server.go
+++ b/cmd/cdc/server/server.go
@@ -143,7 +143,7 @@ func (o *options) run(cmd *cobra.Command) error {
 	select {
 	case <-ch:
 	case <-ticker.C:
-		log.Warn("graceful close timeout")
+		log.Warn("graceful shutdown timeout, exit server")
 	}
 	return err
 }


cmd/cdc/server/server.go
```diff
@@ -132,7 +133,18 @@ func (o *options) run(cmd *cobra.Command) error {
...
+	ticker := time.NewTicker(server.GracefulShutdownTimeout)
+	defer ticker.Stop()
+	go func() {
+		svr.Close(ctx)
+		close(ch)
+	}()
+	select {
+	case <-ch:
+	case <-ticker.C:
+		log.Warn("graceful close timeout")
```
asddongmen:
```suggestion
		log.Warn("graceful shutdown timeout, exit server")
```

Issue #758, opened 2025-01-02
Enhance Unit Test Coverage: Tracking Issue


asddongmen:
# Background
The current project’s unit test coverage has not met the expected target, with insufficient test coverage for some critical modules and edge cases. To improve code quality and system reliability, we need to systematically enhance unit test coverage.

# Goals

- Increase unit test coverage to over 80%.

- Ensure that all core functionalities and edge cases are covered by corresponding test cases.

- Identify and supplement missing tests in the existing codebase.

# Packages 

1. dynstream #756

PR #2789, opened 2025-10-23
*: add more unit tests

wk989898:

Issue Number: ref #758

#### Tests 

- Unit test
- Integration test

### Release note 

```release-note
`None`.
```


pkg/orchestrator/reactor_state_test.go
tenfyzhong: Why don't we delete this file?

PR #2884, opened 2025-10-29
fix(ddl): ensure physical table ID is used for partitioned tables in DDL events

### LINKED ISSUE #2219: Wrong table ID used in MQ open protocol/simple protocol output for partitioned tables
Currently, when TiCDC outputs data to MQ downstream using open protocol or simple protocol, the message includes the logical table ID for partitioned tables instead of the physical table ID.

This causes inconsistencies, since downstream consumers expect the physical table ID to correctly distinguish partitions.

After review, we found that in our codebase, all TableInfo representing a physical table (rather than a logical table) should carry the physical table ID. However, this is not handled properly right now.

To fix this, we should update the following places to ensure the physical table ID is set correctly:
	•	persist_storage_ddl_handlers.go:
		•	extractTableInfoFunc
		•	buildDDLEventFunc

Expected Behavior:
	•	All TableInfo instances representing physical tables should have their physical table ID set.
	•	Output to MQ downstream (open protocol / simple protocol) should contain the correct physical table ID.

Impact:
	•	Downstream consumers may misinterpret partitioned table events because of incorrect table ID usage.

Next Steps:
	•	Update the logic in extractTableInfoFunc and buildDDLEventFunc to ensure physical table IDs are always correctly set for physical tables.

wlwilliamx:

Issue Number: close #2219

This PR refactors how DDL events are created and handled to ensure the physical partition ID is used whenever an event pertains to a specific physical partition.

The central idea is to make `TableInfo.TableName.TableID` the single source of truth for the table ID associated with a DDL event.

1.  **Remove `DDLEvent.TableID`:** The redundant `TableID` field on the `commonEvent.DDLEvent` struct has been removed. This field was ambiguous (logical vs. physical) and often conflicted with the ID stored inside `DDLEvent.TableInfo`. All logic now relies on `DDLEvent.TableInfo.TableName.TableID`.

2.  **Introduce `WrapTableInfoWithTableID`:** A new function, `common.WrapTableInfoWithTableID`, has been added. This function allows creating a `TableInfo` wrapper with an *explicitly* provided `tableID`. The existing `common.WrapTableInfo` function is now a wrapper around this new function, defaulting to the logical table ID (`info.ID`).

3.  **Update DDL Handlers:** The logic in `logservice/schemastore/persist_storage_ddl_handlers.go` has been significantly updated to use this new mechanism:
    * All `extractTableInfoFuncFor...` functions (e.g., `extractTableInfoFuncForAddPartition`) are modified to use `WrapTableInfoWithTableID`, passing in the specific physical partition ID (`tableID`) they are extracting.
    * The DDL builder function `buildDDLEventCommon` has been refactored. A new `buildDDLEventCommonWithTableID` function is now the core builder, which accepts a `tableID` and uses it to create the `TableInfo` wrapper.
    * All DDL builders for partition-specific operations (e.g., `buildDDLEventForAddPartition`, `buildDDLEventForTruncateAndReorganizePartition`, `buildDDLEventForAlterTablePartitioning`) are updated to call `buildDDLEventCommonWithTableID` with the correct physical partition ID.

#### Tests 

- Unit test
- Integration test

None
None
### Release note 

```release-note
None
```

### DIFF
--- a/pkg/sink/codec/open/encoder.go
+++ b/pkg/sink/codec/open/encoder.go
@@ -235,7 +235,11 @@ func enhancedKeyValue(key, value []byte) ([]byte, []byte) {
 
 func (d *batchEncoder) EncodeDDLEvent(e *commonEvent.DDLEvent) (*common.Message, error) {
 	lock.Lock()
-	delete(columnFlagsCache, e.TableInfo.TableName.TableID)
+	tableID := int64(0)
+	if e.TableInfo != nil {
+		tableID = e.TableInfo.TableName.TableID
+	}
+	delete(columnFlagsCache, tableID)
 	defer lock.Unlock()
 
 	key, value, err := encodeDDLEvent(e, d.config)


downstreamadapter/dispatcher/basic_dispatcher.go
```diff
@@ -458,10 +458,14 @@ func (d *BasicDispatcher) handleEvents(dispatcherEvents []DispatcherEvent, wakeC
 				}
 			}
 
+			tableID := int64(0)
```
wk989898:
```suggestion
			tableID := int64(-1)
```

wlwilliamx: 0 means a schema ddl

wlwilliamx: we can't use -1 here

PR #3094, opened 2025-11-13
azure: support managed identity and fix a goroutine hang issue

### LINKED ISSUE #3093: make TiCDC support managed identity for Azure
Today things like blob store access on azure is controlled by authentication via fixed azure keys/secrets as mentioned in: [Grant Permissions to Remote Storage](https://docs.pingcap.com/tidb-in-kubernetes/stable/grant-permissions-to-remote-storage#azure-account-permissions)

This approach is less secure than managed identities which is managed and provisioned by azure gated by the vm identity and we would like to switch to using that: [Managed identities for Azure resources - Managed identities for Azure resources](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview)

Currently this not appear to be supported and we would like to make a feature request to have this supported.

Notably instead of using a k8s secret (that is fixed and requires us to manage/rotate), I believe this feature would require binaries (like tikv) to query the local endpoint to obtain the credentials

e.g. [How managed identities for Azure resources work with Azure virtual machines - Managed identities for Azure resources](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-managed-identities-work-vm)

curl 'http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https%3A%2F%2Fstorage.azure.com%2F' -H Metadata:true

wlwilliamx:

Issue Number: close #3093

TiCDC writes directly to Azure Blob Storage for the azblob sink. This PR adds Azure Managed Identity / Workload Identity token auth support and fixes a potential goroutine hang in cloud storage uploads.

- Support Azure token-based auth (Managed Identity / Workload Identity) for azblob sinks.
- Add timeout/cancellation handling for streaming/multipart uploads to avoid stuck goroutines.

#### Tests

- Unit test
- Manual test (E2E on Azure)

Manual test steps:
1. Deploy TiCDC and TiDB cluster on Azure with Managed Identity / Workload Identity enabled and configure an azblob sink pointing to Azure Blob Storage.
2. Run a changefeed and generate writes.
3. Verify objects are uploaded to the target container/prefix and the changefeed continues running over time.

No expected regression; existing SAS-based configs remain supported.

N/A

### Release note

```release-note
Support Azure Managed Identity authentication for azblob sinks and avoid potential hangs during cloud storage uploads.
```


lidezhu (approved):

pkg/util/external_storage.go
```diff
@@ -69,14 +69,14 @@ func getExternalStorage(
 	})
 	if err != nil {
 		retErr := errors.ErrFailToCreateExternalStorage.Wrap(errors.Trace(err))
```
3AceShowHand: just call errors.WrapError(ErrFailToCreateExternalStorage, err) is the correct usage

pkg/util/external_storage.go
```diff
@@ -244,21 +244,104 @@ func (s *extStorageWithTimeout) WalkDir(
...
+		ctx, cancelCreate = context.WithCancel(ctx)
 	}
-	// multipart uploading spawns a background goroutine, can't set timeout
+
 	writer, err := s.ExternalStorage.Create(ctx, path, option)
 	if err != nil {
+		if cancelCreate != nil {
+			cancelCreate()
+		}
 		err = errors.ErrExternalStorageAPI.Wrap(err).GenWithStackByArgs("Create")
```
3AceShowHand: ditto

pkg/util/external_storage.go
```diff
@@ -244,21 +244,104 @@ func (s *extStorageWithTimeout) WalkDir(
...
+	err := w.ExternalFileWriter.Close(ctx)
+
+	if stop != nil {
+		stop()
+	}
+	if cancel != nil {
+		cancel()
+	}
+	if err != nil {
+		err = errors.ErrExternalStorageAPI.Wrap(err).GenWithStackByArgs("Close")
```
3AceShowHand: ditto

PR #3199, opened 2025-11-18
fix(filter): fix schema DDL filtering bug and refactor filter interface

### LINKED ISSUE #3198: Schema-level DDLs bypass the event filter because TableInfo is nil
Currently, schema-level DDLs (such as `CREATE DATABASE` or `DROP DATABASE`) are not correctly processed by the filter module. These events typically have an empty table name and a `nil` TableInfo.

**Root Cause**
In the `persist_storage_ddl_handlers.go` logic, there is a pre-check condition:
`if tableFilter != nil && rawEvent.SchemaName != "" && rawEvent.TableName != ""`

Because schema-level DDLs have an empty `TableName`, they fail this check. Consequently, they bypass the filter logic entirely and are replicated downstream, even if the user has configured rules to ignore them. Furthermore, the `ShouldIgnoreDDL` interface expects a `TableInfo` object, which causes issues when handling these schema-only events where no table information exists.

**Expected Behavior**
The filter module should evaluate all DDLs, including those that only affect the schema (database), to determine if they should be replicated or ignored.

wlwilliamx:

Issue Number: close #3198

This PR addresses a bug where schema-level DDL events (like `CREATE DATABASE`) were not being passed through the filter module correctly, as well as significantly refactoring the DDL/table filtering API to simplify its usage and reduce reliance on often-unavailable metadata.

#### **Key Changes & Bug Fixes:**

1.  **Bug Fix: Schema DDL Filtering:**
    * Previously, in `logservice/schemastore/persist_storage_ddl_handlers.go`, the DDL filtering logic for schema/table DDLs was guarded by an explicit check that the schema and table names were non-empty (or implicitly by passing `nil` for `tableInfo` which was then checked inside the filter).
    * This logic caused schema DDLs (where the table name is often empty) to bypass filtering entirely.
    * The PR modifies functions like `filterDDL` to call the filter check (`tableFilter.ShouldDiscardDDL` or `tableFilter.ShouldIgnoreDDL`) regardless of whether the schema or table name is empty, ensuring that DDLs are filtered based on the available schema/table name pair.

2.  **Filter API Refactoring:**
    * The `pkg/filter.Filter` interface methods `ShouldIgnoreTable` and `ShouldIgnoreDDL` have been simplified.
    * **Removed the `*common.TableInfo` parameter** from the following functions:
        * `Filter.ShouldIgnoreTable(schema, table string)`
        * `Filter.ShouldIgnoreDDL(schema, table, query string, ddlType timodel.ActionType, startTs uint64)`
    * This removes the unnecessary dependency on obtaining detailed table metadata before checking if a table or DDL event should be ignored. The primary filtering logic (which is based on allow/deny lists using schema and table names) does not require the full `TableInfo` object.

3.  **Widespread Call Site Updates:**
    * Updated numerous call sites across the codebase (e.g., `cmd/filter-helper`, `logservice/schemastore`, `logservice/validator`, `pkg/filter`) to match the new, simplified function signatures.
    * A null check was added in `common.WrapTableInfo` to safely return `nil` if the input `*model.TableInfo` is `nil`, although the core purpose of this PR is to make that check less relevant to the filtering decision itself.

4.  **New Integration Test:**
    * Added a new integration test to `tests/integration_tests/event_filter/` that explicitly verifies two scenarios:
        * A database that is not in the filter rules is **not** replicated downstream.
        * A table in a filtered database that is **not** in the allow list is **not** replicated downstream, while an allowed table in the same database **is** replicated. This confirms both schema and table-level filtering are working as expected.

These changes clean up the filtering API and, more importantly, ensure that schema-level DDLs are correctly filtered according to user-defined rules.
#### Tests 

- Unit test
- Integration test
- Manual test

None
None
### Release note 

```release-note
None
```

### DIFF
--- a/logservice/schemastore/persist_storage_test.go
+++ b/logservice/schemastore/persist_storage_test.go
@@ -66,6 +66,131 @@ func TestApplyDDLJobs(t *testing.T) {
 		fetchTableDDLEventsTestCase []FetchTableDDLEventsTestCase        // test cases for fetchTableDDLEvents, nil means not check it
 		fetchTableTriggerDDLEvents  []FetchTableTriggerDDLEventsTestCase //	test cases for fetchTableTriggerDDLEvents, nil means not check it
 	}{
+		// test filter rules
+		{
+			testName:       "filter_rules",
+			initailDBInfos: nil,
+			ddlJobs: []*model.Job{
+				buildCreateSchemaJobForTest(100, "filter", 1000),
+				buildCreateTableJobForTest(100, 200, "t1", 1010), // create table 200
+				buildCreateTableJobForTest(100, 201, "t2", 1020), // create table 201
+				buildCreateSchemaJobForTest(101, "haha", 2000),
+				buildCreateTableJobForTest(101, 202, "t2", 2010), // create table 202
+			},
+			tableMap: map[int64]*BasicTableInfo{
+				200: {
+					SchemaID: 100,
+					Name:     "t1",
+				},
+				201: {
+					SchemaID: 100,
+					Name:     "t2",
+				},
+				202: {
+					SchemaID: 101,
+					Name:     "t2",
+				},
+			},
+			partitionMap: map[int64]BasicPartitionInfo{},
+			databaseMap: map[int64]*BasicDatabaseInfo{
+				100: {
+					Name: "filter",
+					Tables: map[int64]bool{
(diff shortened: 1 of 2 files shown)


wlwilliamx:
> Could you add a unit test for the create database DDL?

@wk989898 done

Issue #3563, opened 2025-12-10
CDC panic due to invalid handleColID configuration → decoding failed.


asddongmen:
```
panic: not supported in IntHandle [recovered]

2025-12-10 16:11:49	
[2025/12/10 08:11:49.907 +00:00] [ERROR] [event_scanner.go:705] ["panic when appending row to DML event"] [panic="not supported in IntHandle"] [shouldSplitTxn=true] [dmlEventMaxRows=256] [dmlEventMaxBytes=1048576] [batchDMLEventCount=1] [batchDMLEventLength=0] [handleColIDs="[24,28]"] [recordID=7493989780260962481] [recordIDIsInt=true] [rawEventOpType=2] [rawEventCommitTs=462763258656065521] [rawEventStartTs=462763258642957937] [rawEventRegionID=16118225] [rawEventKeyLen=0] [rawEventValueLen=0] [rawEventOldValueLen=0] [rawEventKeyBytes=19] [rawEventValueBytes=0] [rawEventOldValueBytes=237] [rawEventIsUpdate=false] [rawEventIsDelete=true] [dispatcherID=1320556945934769629010888967281572168633] [physicalTableID=287096] [dmlStartTs=462763258642957937] [dmlCommitTs=462763258656065521] [currentDMLEventRows=0] [currentDMLEventSize=0]
```

PR #3564, opened 2025-12-10
tableInfo: Fix wrong setting for HandleColID

hongyunyan:

Issue Number: close https://github.com/pingcap/ticdc/issues/3563

This pull request resolves an issue where the `HandleColID` field was ambiguously used for both the logical handle key and the row format decoder's handle. By introducing a dedicated `HandleKeyIDList` for the ordered handle key column IDs and clarifying the specific role of `HandleColID`, this change ensures that the system accurately identifies and retrieves the correct handle key for tables, improving data consistency and simplifying related internal logic.

### Highlights

* **Refactored Handle Key Management**: Introduced a new field, `HandleKeyIDList`, within the `columnSchema` struct to explicitly store the ordered list of column IDs that constitute the table's handle key (either the primary key or a selected unique key).
* **Clarified `HandleColID` Usage**: The existing `HandleColID` field has been re-purposed and its documentation updated to indicate it is now solely used for the new row format decoder, specifically to hold the TiDB `Handle` (which can be `-1` for non-clustered tables), separating it from the logical handle key identification.
* **Simplified Handle Key Retrieval**: The `GetOrderedHandleKeyColumnIDs` function in `table_info.go` has been simplified to directly return the `HandleKeyIDList`, removing complex conditional logic that previously attempted to derive the handle key from `PKIndex` and `HandleColID`.
* **Improved Handle Key Initialization Logic**: The logic in `newColumnSchema` and `initIndexColumns` was updated to correctly populate `HandleKeyIDList` based on whether a primary key or a suitable unique key exists, ensuring accurate handle key identification.
* **Enhanced Test Coverage**: New test cases were added to `table_info_helper_test.go` to validate the correct behavior of `GetOrderedHandleKeyColumnIDs` under various index configurations, including unclustered indexes and tables with unique keys.

#### Tests 

- Unit test
- Integration test
- No code

### Release note 

```release-note

```

### DIFF
--- a/pkg/common/table_info_helper.go
+++ b/pkg/common/table_info_helper.go
@@ -361,7 +361,7 @@ type columnSchema struct {
 
 	// TiCDC only replicates table has primary key or not null unique key.
 	// * If there is primary key, it's the handle key
-	// * Otherwise, it's the unique key. If there are multiple unique keys, it's the one with the longest length
+	// * Otherwise, it's the unique key. If there are multiple unique keys, it's the longest one.
 	// * If there are multiple unique key with the same length, it the first one.
 	HandleKeyIDs map[int64]struct{} `json:"handle_key_ids"`
 	// HandleKeyIDList store the colID list of the HandleKeyIDs in order.


pkg/common/table_info_helper.go
```diff
@@ -641,17 +653,16 @@ func (s *columnSchema) initIndexColumns() {
 		}
 	}
 
-	if handleIndexOffset < 0 {
+	if handleIndexOffset < 0 || len(s.HandleKeyIDs) > 0 {
```
3AceShowHand: It's better add some comment here, it looks this code implies something not obvious.

PR #3863, opened 2025-12-31
fix(ddl): wait running add index before CREATE TABLE LIKE

### LINKED ISSUE #3832: fix `create table like` in schemastore
### What did you do?

For the DDL statement create table a like b, we need to add both tables a and b to the DDLEvent.BlockedTables. Otherwise, if there is another DDL operation like add index to b that occurs before the create table, the create table like will not wait for the index addition to finish. Consequently, table a will be created without the expected index.

_No response_

### What did you see instead?

None

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```

wlwilliamx:

Issue Number: close #3832

Fix downstream schema inconsistency for CREATE TABLE ... LIKE ... when the referenced table has a long-running ADD INDEX that is still running/queueing in downstream TiDB. In this case, CREATE TABLE b LIKE a may copy the current downstream schema of a before the index becomes visible, causing b to miss the index.

This PR makes CREATE TABLE ... LIKE ... wait for any running async ADD INDEX on the referenced table before executing the CREATE TABLE ... LIKE ....

  - Detect CREATE TABLE ... LIKE ... in schemastore when building DDLEvent, extract the referenced table, and populate BlockedTableNames with that referenced table.
  - Adjust mysql/TiDB sink behavior so CREATE TABLE is eligible for async-DDL waiting logic; for normal create table (no LIKE), this is a no-op because BlockedTableNames is empty. For CREATE TABLE ... LIKE ..., it triggers waiting on the referenced table’s running ADD INDEX job.

#### Tests 

- Unit test
- Integration test
- Manual test

None
None
### Release note 

```release-note

```

### DIFF
--- a/logservice/schemastore/persist_storage_ddl_handlers.go
+++ b/logservice/schemastore/persist_storage_ddl_handlers.go
@@ -516,18 +516,20 @@ func getSchemaID(tableMap map[int64]*BasicTableInfo, tableID int64) int64 {
 	return tableInfo.SchemaID
 }
 
+// schemaName should be "Name.O"
 func findSchemaIDByName(databaseMap map[int64]*BasicDatabaseInfo, schemaName string) (int64, bool) {
 	for id, info := range databaseMap {
-		if strings.EqualFold(info.Name, schemaName) {
+		if info.Name == schemaName {
 			return id, true
 		}
 	}
 	return 0, false
 }
 
+// tableName should be "Name.O"
 func findTableIDByName(tableMap map[int64]*BasicTableInfo, schemaID int64, tableName string) (int64, bool) {
 	for id, info := range tableMap {
-		if info.SchemaID == schemaID && strings.EqualFold(info.Name, tableName) {
+		if info.SchemaID == schemaID && info.Name == tableName {
 			return id, true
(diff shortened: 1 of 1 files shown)


pkg/sink/mysql/helper.go
```diff
@@ -399,8 +399,6 @@ func needSwitchDB(event *commonEvent.DDLEvent) bool {
 
 func needWaitAsyncExecDone(t timodel.ActionType) bool {
 	switch t {
-	case timodel.ActionCreateTable, timodel.ActionCreateTables:
```
wk989898: I think it is not a good idea to remove this. If there are lots of create table DDL, every table will query the downstream. It may affect performance. Could we handle the `create table like` separately?

wlwilliamx: Removing `create table` and `create tables` here is irrelevant, because only `create table like` will have blocked table names. Only when there are blocked table names will the query `add index` and wait.

hongyunyan: Why we could remove this function?

PR #3956, opened 2026-01-08
feat(ticdc): prevent using the same TiDB cluster as both upstream and downstream

### LINKED ISSUE #3949: Prevent TiCDC from Using the Same TiDB Cluster as Both Upstream and Downstream
## Background  

TiCDC allows replicating data from an upstream TiDB cluster to a downstream system, including another TiDB cluster. However, using the same TiDB cluster as both the upstream and downstream can cause unexpected replication issues, including feedback loops and data inconsistencies.  

To prevent this, we need a mechanism to reliably determine whether the upstream and downstream TiDB clusters are the same and reject such configurations.  

## Proposed Solution  

We will use the **Cluster ID** stored in TiDB’s system table(`mysql.tidb`) to compare the upstream and downstream clusters.  

1. **Retrieve Upstream Cluster ID**  
   - TiCDC already obtains the upstream Cluster ID via PD’s gRPC API.  

2. **Retrieve Downstream Cluster ID**  
   - TiDB will introduce a system table  `mysql.tidb` to store the Cluster ID.  
   - TiCDC will query this system table via SQL to fetch the downstream Cluster ID.  

3. **Comparison and Validation**  
   - If the upstream and downstream Cluster IDs match, TiCDC will reject the changefeed creating, updating and resuming with an appropriate error message.  

## Tasks  

- [x] Implement the system table in TiDB to store the Cluster ID.  [tidb#59511](https://github.com/pingcap/tidb/pull/59511)
- [ ] Modify TiCDC to query the `mysql.tidb` and retrieve the downstream Cluster ID.  
- [ ] Compare the upstream and downstream Cluster IDs during changefeed creation.  
- [ ] Add error handling and logging when the same Cluster ID is detected.

wlwilliamx:
  TiCDC does not support using the same TiDB logical cluster as both the upstream and downstream, because it can easily lead to self-replication loops or unexpected data behaviors.

  In TiDB Next-Gen, a physical cluster can contain multiple logical clusters (keyspaces). Different keyspaces should be treated as different clusters from TiCDC's perspective. So this PR keeps rejecting same upstream/downstream for the same keyspace, while allowing cross-keyspace replication within the same physical cluster.

Issue Number: close #3949

  - Add a pre-check for changefeed create/update/resume to detect “same upstream & downstream” for MySQL-compatible sinks.
    - The downstream TiDB cluster ID is read from `mysql.tidb` (`cluster_id`).
    - If downstream is not TiDB or the cluster ID is unavailable, the check is skipped (legacy/compat behavior).
  - Next-Gen compatibility: when the upstream and downstream `cluster_id` are the same, query the downstream keyspace via:
    - `show config where type = 'tidb' and name = 'keyspace-name'`
    - Treat `(cluster_id, keyspace)` as the cluster identity, so cross-keyspace is allowed but same keyspace is rejected.
    - If downstream keyspace cannot be determined, keep the conservative legacy behavior (treat as the same physical cluster and reject) to avoid accidental self-replication.
  - Add unit tests and an integration test case `same_upstream_downstream`.

  ### Check List 

  #### Tests 

  - Unit test
  - Integration test
  - Manual test

  #### Questions 

  - No steady-state overhead: the check only runs when creating/updating/resuming a changefeed.
  - Classic mode keeps the existing behavior (same physical cluster is rejected).
  - Next-Gen allows cross-keyspace replication within the same physical cluster, while still rejecting same keyspace.

  N/A

  ### Release note 

```release-note
 Reject creating/updating/resuming a changefeed when downstream is the same TiDB logical cluster as upstream (keyspace-aware for TiDB Next-Gen).
 ```

### CODE REVIEW

**Comment by wlwilliamx** [comment]
CC @tenfyzhong

**Comment by wlwilliamx** [comment]
CC @asddongmen

File: api/v2/changefeed.go [hunk 1]
```diff
@@ -249,6 +250,18 @@ func (h *OpenAPIV2) CreateChangefeed(c *gin.Context) {
 
 	// verify sinkURI
 	cfConfig := info.ToChangefeedConfig()
+	// Check whether the upstream and downstream are the different cluster.
+	notSame, err := check.UpstreamDownstreamNotSame(ctx, pdClient, cfConfig)
```
tenfyzhong: **Confusing Naming and Logic**: The function name `UpstreamDownstreamNotSame` with double negative logic (`!notSame`) makes the code harder to understand.

Consider renaming to use positive semantics for better readability:

```go
// Rename to IsSameCluster
isSame, err := check.IsSameCluster(ctx, pdClient, cfConfig)
if err != nil {
    _ = c.Error(err)
    return
}
if isSame {
    _ = c.Error(errors.ErrSameUpstreamDownstream.GenWithStack(...))
    return
}
```

Either naming would be clearer than the current `notSame` variable combined with `!notSame` check. The current double negative makes the logic harder to follow.

pkg/check/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+			continue
+		}
+		if value != keyspace {
+			return "", cerrors.New("downstream TiDB reports inconsistent keyspace-name across instances")
+		}
+	}
+	if err := rows.Err(); err != nil {
+		return "", cerrors.Trace(err)
+	}
+	if !hasRow {
```
tenfyzhong: **Classic Architecture Compatibility**: When no `keyspace-name` is found, we should return `common.DefaultKeyspaceName` instead of returning an error. This ensures compatibility with TiDB Classic architecture.

In TiDB Classic (non-Next-Gen), the `SHOW CONFIG` command may not return the `keyspace-name` row at all, or the configuration may not exist. Returning an error here would break the check for Classic deployments.

Suggested fix:
```go
if !hasRow {
    // TiDB Classic doesn't expose keyspace-name, use default
    return common.DefaultKeyspaceName, nil
}
```

api/v2/changefeed.go
wlwilliamx: Agreed that `UpstreamDownstreamNotSame` + `!notSame` was confusing. I refactored it to positive semantics:
- renamed to `check.IsSameUpstreamDownstream`
- updated API handlers to use `isSame` and `if isSame { ... }`
This should make the intent easier to follow.

pkg/check/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+	for rows.Next() {
+		// Columns: Type | Instance | Name | Value
+		var typ, instance, name, value string
+		if err := rows.Scan(&typ, &instance, &name, &value); err != nil {
+			return "", cerrors.Trace(err)
+		}
+		if !hasRow {
+			keyspace = value
+			hasRow = true
+			continue
```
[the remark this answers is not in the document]
wlwilliamx: I kept iterating all rows intentionally to validate `keyspace-name` consistency across TiDB instances. This check only runs on create/update/resume, so the overhead is negligible, and breaking early would skip the consistency guard.

PR #4001, opened 2026-01-14
dispatcher: fix pendingACKCount leak on block event write error

### LINKED ISSUE #4000: Table-trigger dispatcher may leak pendingACKCount on sink write error, potentially stalling DB/All block events
### What did you do?

A potential `pendingACKCount` leak exists on an error path in `BasicDispatcher.DealWithBlockEvent`.

For schedule-related non-blocking DDLs (i.e. a single-table DDL that sets `NeedAddedTables` or `NeedDroppedTables`), `pendingACKCount` is incremented before writing the block event downstream. If the downstream write fails (e.g. `sink.WriteBlockEvent` returns an error), the function returns early before creating a resend task, which means the maintainer will never ACK this event.

`pendingACKCount` should only track events that will eventually receive a maintainer ACK. If writing the block event to downstream fails before a resend task is created, the counter should be rolled back to avoid blocking unrelated events.

### What did you see instead?

On the error path described above, `pendingACKCount` can remain > 0 indefinitely.

Later DB/All block events (e.g. syncpoint or DB-level DDL) may be held because the table-trigger dispatcher uses `pendingACKCount > 0` as the condition to hold DB/All events until all schedule-related ACKs are done, potentially stalling progress.

### Versions of the cluster

```console
N/A (code reading finding; not tied to a specific cluster version)
```

```console
N/A
```

```console
master (observed on commit e78fdb22ce2bdc3c4474d0b68ded8dda8d70e836)
```

wlwilliamx:

Issue Number: close #4000

Roll back `pendingACKCount` when writing a schedule-related non-blocking block event to downstream fails, so DB/All block events won't be held forever due to a counter leak.

#### Tests 

- Unit test
- Integration test

No. This only affects the error path when writing a schedule-related non-blocking DDL fails before a resend task is created.

No.

### Release note 

```release-note
Fix a pendingACKCount leak in the table-trigger dispatcher that could stall DB/All block events after a schedule-related DDL write error.
```


downstreamadapter/dispatcher/basic_dispatcher.go
```diff
@@ -728,17 +728,22 @@ func (d *BasicDispatcher) DealWithBlockEvent(event commonEvent.BlockEvent) {
...
+			needsScheduleACKTracking := d.IsTableTriggerEventDispatcher() && !noNeedAddAndDrop
+			if needsScheduleACKTracking {
+				// If this is a table trigger event dispatcher, and the DDL leads to add/drop tables,
+				// we track it as a pending schedule-related event until the maintainer ACKs it.
 				d.pendingACKCount.Add(1)
 			}
 			err := d.AddBlockEventToSink(event)
 			if err != nil {
+				if needsScheduleACKTracking {
+					d.pendingACKCount.Add(-1)
```
tenfyzhong: Good fix! The rollback mechanism correctly addresses the pendingACKCount leak issue. 

However, I'd like to understand the timing better: According to the [Block Event Handling documentation](https://deepwiki.com/pingcap/ticdc/6.5-block-event-handling-in-dispatchers), when `AddBlockEventToSink` fails, does this mean:
1. The block event was never sent to the sink at all, OR
2. The sink write started but failed midway?

If it's case 2, could there be any partial state in the sink that needs cleanup? The current implementation seems to assume case 1 is always true, which is likely correct for this error path (before resend task creation), but a comment clarifying this assumption would be helpful.

PR #4031, opened 2026-01-21
message: send request to get table info from schemastore

### LINKED ISSUE #2985: Adjusting the logic for emitting the table bootstrap event
The event dispatcher can't use the schemastore. We have to send messages to schemastore and wait for the messages to get all tableinfos

wk989898:

Issue Number: close #2985

 - Introduces a shared schemastoreclient for fetching table schemas, registering keyspaces, and listing physical tables. Dispatcher bootstrap, API, coordinator, and maintainer use this client.

  - Adds SchemaStore message handlers in logservice. For bootstrap, the client sends table IDs and a timestamp; the server returns each table's schema or error, followed by a completion response.

  - Adds request tracking, cancellation, timeouts, send retries, and completeness checks so missing responses cannot silently produce an incomplete bootstrap. Tests and mocks are updated to cover the new message flow.

#### Tests 

- Unit test
- Integration test

### Release note 

```release-note
`None`.
```

## Summary by CodeRabbit

* **Improvements**
  * Schema and table metadata requests now use batched retrieval for more efficient bootstrap and replication processing.
  * Added request timeouts, cancellation handling, response validation, and clearer error reporting.
  * Schema-store operations now better handle queue limits, unavailable data, snapshot expiration, and service shutdown.
  * Bootstrap processing stops safely when required table metadata cannot be retrieved.
* **Bug Fixes**
  * Improved handling of duplicate, delayed, malformed, or partial schema-store responses.
  * Canceled metadata requests now terminate promptly without leaving pending operations behind.

### DIFF
--- a/logservice/schemastore/message_handler.go
+++ b/logservice/schemastore/message_handler.go
@@ -1,0 +1,133 @@
+// Copyright 2025 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schemastore
+
+import (
+	"context"
+
+	"github.com/pingcap/log"
+	"github.com/pingcap/ticdc/pkg/common"
+	"github.com/pingcap/ticdc/pkg/messaging"
+	"github.com/pingcap/ticdc/pkg/node"
+	"go.uber.org/zap"
+)
+
+func (s *schemaStore) handleMessage(ctx context.Context, msg *messaging.TargetMessage) error {
+	switch msg.Type {
+	case messaging.TypeSchemaStoreTableInfosRequest:
+		for _, m := range msg.Message {
+			req, ok := m.(*messaging.SchemaStoreTableInfosRequest)
+			if !ok {
+				log.Warn("invalid schema store request message, ignore it",
+					zap.Any("msg", msg))
+				continue
+			}
+			from := msg.From
+			go s.handleTableInfosRequest(ctx, from, req)
+		}
+	default:
+		log.Warn("unknown message type, ignore it",
+			zap.String("type", msg.Type.String()),
+			zap.Any("message", msg.Message))
+	}
+	return nil
+}
+
+func (s *schemaStore) handleTableInfosRequest(
+	ctx context.Context,
+	from node.ID,
+	req *messaging.SchemaStoreTableInfosRequest,
+) {
+	if req == nil {
+		return
+	}
+
+	sendResponse := func(resp *messaging.SchemaStoreTableInfosResponse) {
+		if resp == nil {
+			return
+		}
+		msg := messaging.NewSingleTargetMessage(from, messaging.SchemaStoreClientTopic, resp)
+		err := s.mc.SendCommand(msg)
+		if err != nil {
+			log.Warn("send schema store response failed",
+				zap.Any("keyspaceID", req.KeyspaceID),
+				zap.Uint64("requestID", req.RequestID),
+				zap.Int64("tableID", resp.TableID),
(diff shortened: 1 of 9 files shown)


3AceShowHand (approved):

logservice/schemastore/message_handler.go
```diff
@@ -0,0 +1,309 @@
...
+	message   *messaging.SchemaStoreRequest
+	expiresAt time.Time
+}
+
+type schemaRequestSubmission struct {
+	request schemaRequest
+	result  chan error
+}
+
+func newSchemaStoreMessageHandler(ctx context.Context, store *schemaStore, mc messaging.MessageCenter, workers int) *schemaStoreMessageHandler {
```
lidezhu, line 52: This design is too complex. I think we can just use a simple solution. For example:
```
  type schemaStoreMessageHandler struct {
      ctx      context.Context
      cancel   context.CancelFunc
      store    *schemaStore
      mc       messaging.MessageCenter
      requests chan schemaRequest
      wg       sync.WaitGroup
  }

  func (h *schemaStoreMessageHandler) submit(req schemaRequest) error {
      if err := h.ctx.Err(); err != nil {
          return err
      }

      select {
      case h.requests <- req:
          return nil
      default:
          return errors.ErrSchemaStoreRequestFailed.
              GenWithStack("schema store request queue is full")
      }
  }

  func (h *schemaStoreMessageHandler) runWorker() {
      defer h.wg.Done()

      for {
          select {
          case <-h.ctx.Done():
              return
          case req := <-h.requests:
              ctx, cancel := context.WithTimeout(
                  h.ctx,
                  messaging.SchemaStoreRequestTimeout,
              )
              h.handleRequest(ctx, req.from, req.message)
              cancel()
          }
      }
  }
```

pkg/messaging/schema_store_message.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2026 PingCAP, Inc.
```
lidezhu: SchemaStore messages should live in the SchemaStore package rather than the generic messaging package. Also, for consistency with other cross-component messages, these request and response types should probably be defined using protobuf.

lidezhu:
Remove RegisterKeyspace from this client. Its lifecycle semantics are not settled: the current call only registers the SchemaStore on the local node, even though every node owns a SchemaStore. Sending the same operation through MessageCenter with the target set to the local node does not address that problem; it only encodes the existing single-node behavior into the new protocol. Keep the existing call unchanged for now and design keyspace registration separately.

lidezhu:
Define a separate protobuf request/response pair for each operation. The current operation-based request contains many fields that are invalid or unused for most operations, makes invalid field combinations representable, and couples unrelated APIs to the same wire format. Dispatching by concrete message type also removes the SchemaStoreOperation switch and allows each API to evolve independently.
```
  message GetTableInfosRequest {
    uint64 request_id = 1;
    KeyspaceMeta keyspace = 2;
    repeated int64 table_ids = 3;
    uint64 ts = 4;
  }

  message GetTableInfosResponse {
    uint64 request_id = 1;
    repeated TableInfoResult table_infos = 2;
    bool more = 3;
    Error error = 4;
  }

  message GetAllPhysicalTablesRequest {
    uint64 request_id = 1;
    KeyspaceMeta keyspace = 2;
    uint64 ts = 3;
    FilterConfig filter = 4;
    bool case_sensitive = 5;
    bool force_replicate = 6;
  }

  message GetAllPhysicalTablesResponse {
    uint64 request_id = 1;
    repeated PhysicalTable tables = 2;
    Error error = 3;
  }
```

pkg/schemastore/message.proto
```diff
@@ -0,0 +1,79 @@
...
+    repeated int64 table_ids = 3 [(gogoproto.customname) = "TableIDs"];
+    uint64 ts = 4;
+}
+
+message GetTableInfosResponse {
+    uint64 request_id = 1 [(gogoproto.customname) = "RequestID"];
+    // Results cover a prefix of the requested table_ids, in the same order.
+    repeated TableInfoResult table_infos = 2 [(gogoproto.nullable) = false];
+    // The client must issue a new request for the remaining table_ids.
+    bool more = 3;
```
lidezhu: Seems `more` is not needed?

PR #4060, opened 2026-01-23
coordinator: add gc cleaner to clean up changefeed level service safe point

### LINKED ISSUE #4126: gc cleaner to clean up changefeed level service safe point
The changefeed service safe point is clean up periodically by the gc cleaner

3AceShowHand:

Issue Number: close #4126 

* gc cleaner is responsible for clean up changefeed level service safe point, after the coordinator make sure update gc safe point successed once.

On the current master branch, when creating and resuming changefeed with new start-ts, a changefeed level service safe point is put to the pd, and it will be removed after the changefeed become normal. 

The remove changefeed level service safe point may too early, since the changefeed checkpoint is not take part in the calculation of the cdc cluster level service safe point yet, which is triggered by the `updateGCSafepoint` method.

This PR aims to make sure, after the coordinator call `updateGCSafepoint` successfully, then remove the changefeed level service safe point.

#### Tests 

- Unit test
- Integration test

### Release note 

```release-note
`None`
```

## Summary by CodeRabbit

* **Chores**
  * Improved garbage-collection management for changefeeds with a dedicated cleanup worker, stronger ordering guarantees, and non-blocking safepoint handling.
  * Made GC tick cadence configurable to control GC timing and reliability.
  * Changefeed lifecycle now registers cleanup tasks instead of direct safepoint updates for safer coordination.

* **Tests**
  * Added extensive unit and integration tests covering GC cleanup, async safepoint behavior, and retry/batching scenarios.

* **Chores**
  * Adjusted test runner config to speed GC-ticker-driven tests.

### DIFF
--- a/tests/integration_tests/gc_safepoint/run.sh
+++ b/tests/integration_tests/gc_safepoint/run.sh
@@ -94,7 +94,7 @@ function run() {
 	esac
 	# Speed up the coordinator GC tick interval to make the test finish quickly.
 	# The test checks GC safepoint changes within seconds (see check_safepoint_forward / check_safepoint_equal).
-	export GO_FAILPOINTS='github.com/pingcap/ticdc/coordinator/InjectUpdateGCTickerInterval=return(1000)'
+	export GO_FAILPOINTS='github.com/pingcap/ticdc/coordinator/InjectUpdateGCTickerInterval=return(5)'
 	run_cdc_server --workdir $WORK_DIR --binary $CDC_BINARY --addr "127.0.0.1:8300" --pd $pd_addr
 	changefeed_id=$(cdc_cli_changefeed create --pd=$pd_addr --sink-uri="$SINK_URI" | grep '^ID:' | head -n1 | awk '{print $2}')
 	case $SINK_TYPE in


coordinator/coordinator.go
```diff
@@ -274,22 +282,6 @@ func (c *coordinator) handleStateChange(
 		c.controller.moveChangefeedToSchedulingQueue(event.changefeedID, false, false)
 	case config.StateFailed, config.StateFinished:
 		c.controller.operatorController.StopChangefeed(ctx, event.changefeedID, false)
-	case config.StateNormal:
```
wk989898: Why remove this?

coordinator/coordinator.go
```diff
@@ -363,12 +355,12 @@ func (c *coordinator) saveCheckpointTs(ctx context.Context, changes []*changefee
...
-	err := c.controller.CreateChangefeed(ctx, info)
-	if err != nil {
-		return errors.Trace(err)
+	if err := c.controller.CreateChangefeed(ctx, info); err != nil {
+		return err
 	}
-	// update gc safepoint after create changefeed
-	return c.updateGCSafepointByChangefeed(ctx, info.ChangefeedID)
+
+	c.gcCleaner.Add(info.ChangefeedID, info.KeyspaceID, gc.EnsureGCServiceCreating)
```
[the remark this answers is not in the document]
wk989898: Could you add a failpoint to simulate it?

coordinator/gccleaner/cleaner.go
```diff
@@ -0,0 +1,271 @@
...
+		}
+		if entry.creating && entry.creatingAtEpoch < epoch && len(tasks) < maxTasks {
+			tasks = append(tasks, task{
+				changefeedID: changefeedID,
+				keyspaceID:   entry.keyspaceID,
+				tag:          gc.EnsureGCServiceCreating,
+			})
+			entry.creating = false
+		}
+		if entry.resuming && entry.resumingAtEpoch < epoch && len(tasks) < maxTasks {
```
tenfyzhong: Same redundant check as line 189. The `len(tasks) < maxTasks` condition is unnecessary since we've already checked and broken out of the loop when reaching maxTasks.

coordinator/gccleaner/cleaner.go
```diff
@@ -0,0 +1,271 @@
...
+
+// PendingLen returns the number of changefeeds with pending cleanup tasks.
+func (c *Cleaner) PendingLen() int {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	return len(c.pending)
+}
+
+func (c *Cleaner) Run(ctx context.Context) error {
```
tenfyzhong: **Observation on memory management**: The `pending` map can grow indefinitely if `OnUpdateGCSafepointSucceeded()` is never called (e.g., if GC safepoint update consistently fails). This could be a potential memory leak in pathological scenarios.

Consider adding:
1. A maximum size limit for the `pending` map
2. A TTL mechanism to remove entries that have been pending for too long
3. Metrics to monitor the pending map size

This ensures the cleaner remains robust even when GC updates fail for extended periods.

asddongmen (approved):

PR #4358, opened 2026-03-05
logpuller: fix potential stuck

### LINKED ISSUE #4359: logpuller: subscriptionClient cond pause/resume may miss wakeups
### What did you do?

log puller uses a sync.Cond to block pushRegionEventToDS while the dynamic stream area is paused. The pause/resume state (paused atomic.Bool) was updated and cond.Broadcast() was called without holding the cond’s mutex, which can lead to missed wakeups (a waiter can go to sleep after the broadcast) and result in a goroutine blocking indefinitely.

Additionally, Close() only canceled the context and closed the stream; if the client was paused, goroutines waiting on cond.Wait() had no guarantee to be woken up, risking shutdown hangs/leaks.

_No response_

### What did you see instead?

As described above.

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```

lidezhu:

Issue Number: close #4359 

This pull request addresses a potential deadlock or stuck state within the logpuller's subscription client. It enhances the robustness of the `pushRegionEventToDS` mechanism by integrating context cancellation checks and ensuring proper unblocking during client shutdown, thereby preventing indefinite waits and improving the client's lifecycle management.

### Highlights

* **Prevented potential stuck state in `pushRegionEventToDS`**: The `pushRegionEventToDS` method now checks for context cancellation (`s.ctx.Err()`) before and during its waiting loop, ensuring it can exit gracefully if the context is canceled while the client is paused.
* **Ensured thread safety for `paused` state updates**: Mutex locks (`s.mu.Lock()` and `s.mu.Unlock()`) were added around `s.paused.Store()` operations within `handleDSFeedBack` to prevent race conditions when modifying the client's paused state.
* **Guaranteed unblocking on client closure**: The `Close` method now explicitly sets `s.paused.Store(false)` and broadcasts to the condition variable (`s.cond.Broadcast()`), which ensures any goroutines blocked in `pushRegionEventToDS` are released during shutdown.
* **Added new test case for unblocking on close**: A new unit test, `TestPushRegionEventToDSUnblocksOnClose`, was introduced to verify that the `pushRegionEventToDS` method correctly unblocks when the `subscriptionClient` is closed.

#### Tests 

- Unit test
- Integration test
- No code

### Release note 

```release-note

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Subscriptions now stop promptly when cancelled and unblocks operations on close to avoid hangs.
  * Improved thread-safety around pause/resume transitions to prevent race conditions during subscription handling.

* **Tests**
  * Added tests to verify pause/unpause behavior and that blocked operations are unblocked on close.

### DIFF
--- a/logservice/logpuller/subscription_client.go
+++ b/logservice/logpuller/subscription_client.go
@@ -403,9 +403,6 @@ func (s *subscriptionClient) wakeSubscription(subID SubscriptionID) {
 }
 
 func (s *subscriptionClient) pushRegionEventToDS(subID SubscriptionID, event regionEvent) {
-	if s.ctx.Err() != nil {
-		return
-	}
 	// fast path
 	if !s.paused.Load() {
 		s.ds.Push(subID, event)
@@ -413,13 +410,16 @@ func (s *subscriptionClient) pushRegionEventToDS(subID SubscriptionID, event reg
 	}
 	// slow path: wait until paused is false
 	s.mu.Lock()
-	for s.paused.Load() && s.ctx.Err() == nil {
-		s.cond.Wait()
+	for s.paused.Load() {
+		select {
+		case <-s.ctx.Done():
+			s.mu.Unlock()
+			return
+		default:
+			s.cond.Wait()
+		}
 	}
 	s.mu.Unlock()
-	if s.ctx.Err() != nil {
-		return
-	}
 	s.ds.Push(subID, event)
 }
 


logservice/logpuller/subscription_client.go
```diff
@@ -403,17 +403,23 @@ func (s *subscriptionClient) wakeSubscription(subID SubscriptionID) {
...
+	}
 	// fast path
 	if !s.paused.Load() {
 		s.ds.Push(subID, event)
 		return
 	}
 	// slow path: wait until paused is false
 	s.mu.Lock()
-	for s.paused.Load() {
+	for s.paused.Load() && s.ctx.Err() == nil {
```
wk989898: Why not use `for select` to capture the context error?

PR #4429, opened 2026-03-11
eventBroker: implement two stage syncpoint

### LINKED ISSUE #4605: Syncpoint-enabled throughput drop due to dispatcher memory-quota starvation
## Background

When syncpoint is enabled, each dispatcher independently sends events to EventCollector.
Under DS memory quota pressure, fast dispatchers can continuously fill shared memory,
while slower dispatchers cannot enqueue enough events to make progress.

This creates unfair progress among dispatchers and hurts overall throughput.

## Problem statement

In the current model, EventCollector memory is effectively consumed by whoever sends faster.
There is no global syncpoint-stage coordination in EventBroker to force balanced progress before
moving syncpoint forward.

As a result:

- Fast dispatchers keep advancing and occupy most DS memory quota.
- Slow dispatchers are starved and lag on checkpoint/syncpoint progression.
- Syncpoint completion is delayed by stragglers, causing long tail latency.
- End-to-end throughput drops because the pipeline repeatedly waits on the slowest blocked paths.

## ASCII view of the original problem

```text
Before fix (single-dispatcher-driven pushing)

  Dispatcher A (fast)  ---->  EventCollector queue  [##########]  (quota almost full)
  Dispatcher B (slow)  --x    enqueue blocked / tiny share
  Dispatcher C (slow)  --x    enqueue blocked / tiny share

  Shared DS memory quota:
  +---------------------------------------------------------------+
  | A: 85-95%                                                     |
  | B: 3-8%                                                       |
  | C: 2-7%                                                       |
  +---------------------------------------------------------------+

  Syncpoint view:
    A: sentResolvedTs/checkpointTs keeps moving
    B,C: lagging, cannot catch up
    => global syncpoint commit waits for lagging dispatchers
    => high syncpoint latency + lower throughput
```

## Why this is a performance issue

It is a steady-state performance issue under skewed
workloads: memory quota is monopolized by fast paths, causing starvation and prolonged syncpoint
barriers, which directly reduces throughput and increases p99 latency.

asddongmen:
Issue Number: close #4605 

  When syncpoint is enabled, all dispatchers independently push events to EventCollector. In some workloads, fast
  dispatchers can consume most DS memory quota, while slower dispatchers are starved. This causes unfair progress,
  larger syncpoint lag, and lower overall throughput.

Here are the test results in the Syncpoint + DDL + Redo + Incremental Scan scenario. They are obtained when using the master branch and this PR respectively, allowing for a comparison between the two.

**Master**
image

**This PR**
image

  This PR introduces a two-stage syncpoint workflow in eventBroker to improve fairness and performance under quota
  pressure.

  - Add global syncpoint states in changefeedStatus: syncPointPreparingTs and syncPointInFlightTs.
  - Enter commit stage only after all active dispatchers reach sentResolvedTs >= preparingTs.
  - Finish commit stage only after all active dispatchers move both nextSyncPoint and checkpointTs beyond the in-
    flight ts.
  - Keep same-ts ordering deterministic for (commitTs, isSyncPoint), ensuring DDL is emitted before syncpoint when
    they share commitTs.
  - Increase complete-status heartbeat frequency from 10s to 1s when syncpoint is enabled, so broker gets
    dispatcher progress faster.
  - Tighten same-ts block action matching with isSyncPoint to avoid ambiguity.
  - Add periodic syncpoint stage debug snapshots to expose blocking dispatchers and bottlenecks.

  ### Check List

  #### Tests

  - Unit test
  - Integration test

  #### Questions

  No compatibility change is expected. For non-syncpoint workloads, behavior is unchanged. For syncpoint-enabled
  workloads, this improves progress fairness and reduces starvation-driven latency.

  No.

  ### Release note

  Improve syncpoint-enabled performance by adding two-stage syncpoint coordination in eventBroker, reducing
  dispatcher starvation under DS memory quota pressure.

## Release Notes

* **New Features**
  * Added sync point state management with prepare and commit stage tracking for improved checkpoint coordination
  * Enhanced DDL and sync point event ordering to ensure correct emission sequence

* **Bug Fixes**
  * Fixed event type matching in dispatcher initialization logic
  * Improved dispatcher status synchronization and heartbeat coordination

* **Tests**
  * Added comprehensive test coverage for sync point state transitions, event ordering, and dispatcher behavior

* **Chores**
  * Updated git ignore patterns

### DIFF
--- a/downstreamadapter/dispatcher/basic_dispatcher.go
+++ b/downstreamadapter/dispatcher/basic_dispatcher.go
@@ -749,11 +749,11 @@ func (d *BasicDispatcher) handleEvents(dispatcherEvents []DispatcherEvent, wakeC
 // and return await=true.
 // The status path will be waked up after the action finishes.
 func (d *BasicDispatcher) HandleDispatcherStatus(dispatcherStatus *heartbeatpb.DispatcherStatus) (await bool) {
-	log.Debug("dispatcher handle dispatcher status",
-		zap.String("dispatcherStatus", common.FormatDispatcherStatus(dispatcherStatus)),
-		zap.Stringer("dispatcher", d.id),
-		zap.Any("action", dispatcherStatus.GetAction()),
-		zap.Any("ack", dispatcherStatus.GetAck()))
+	// log.Debug("dispatcher handle dispatcher status",
+	// 	zap.String("dispatcherStatus", common.FormatDispatcherStatus(dispatcherStatus)),
+	// 	zap.Stringer("dispatcher", d.id),
+	// 	zap.Any("action", dispatcherStatus.GetAction()),
+	// 	zap.Any("ack", dispatcherStatus.GetAck()))
 
 	// Step1: deal with the ack info
 	ack := dispatcherStatus.GetAck()


tests/integration_tests/bdr_mode/run.sh
```diff
@@ -37,7 +37,6 @@ function run() {
 
 	run_sql_file $CUR/data/finished.sql ${UP_TIDB_HOST} ${UP_TIDB_PORT}
 
-	# syncpoint table should exists in secondary tidb, but does not exists in primary cluster
```
wk989898: Why remove this?

PR #4437, opened 2026-03-11
kafka: release sarama clients on cleanup and init failures

### LINKED ISSUE #4064: TiCDC Kafka sink: missing topic keeps changefeed in warning (expected) but pegs CPU/memory
[cpu-mem-pprof.zip](https://github.com/user-attachments/files/24846919/cpu-mem-pprof.zip)

## What did you do?

- Set TiCDC downstream sink to Kafka.
- The Kafka cluster does **not** have the target topic created (e.g. broker auto-create topics disabled, or topic intentionally not created).
- Create/start a changefeed that writes to this non-existent Kafka topic.

- The changefeed can stay in `warning` because the topic does not exist (this is expected).
- TiCDC should keep retrying with reasonable backoff and **should not** consume excessive CPU or memory while waiting.

## What did you see instead?

- The changefeed stays in `warning` (expected), **but** TiCDC CPU and memory usage keep increasing rapidly and eventually saturate the machine (high CPU + OOM-like memory growth).

## Profiles

- CPU profile (pprof): `profile.01251428.out` (attached)
- Heap profile (pprof): `heap-01251435.out` (attached)

## Versions of the cluster

```console
N/A (not captured yet)
```

```console
N/A (not captured yet)
```

```console
N/A (not captured yet)
```

wlwilliamx:

Issue Number: close #4064

When a Kafka sink repeatedly enters warning/retry (for example, when the target topic is missing), TiCDC can leave Sarama clients alive across retries.

For producers created from an existing client, Sarama wraps the provided client with a `nopCloserClient`. That means closing the producer alone does not release the original client owned by TiCDC. If the original client is not closed, background goroutines, broker connections, and related caches/metrics can accumulate.

This PR tightens Sarama client cleanup in Kafka sink paths by:

- keeping admin normal-close aligned with Sarama ownership semantics
- closing the Sarama client when `sarama.NewClusterAdminFromClient` fails
- explicitly closing the underlying client in `saramaSyncProducer.Close()`
- closing the Sarama client when `sarama.NewSyncProducerFromClient` fails
- closing the Sarama client when `sarama.NewAsyncProducerFromClient` fails
- adding unit tests for admin and sync producer cleanup behavior

#### Tests 

- Unit test

Unit test:
- `go test ./pkg/sink/kafka -run 'TestSarama(AdminClientCloseDelegatesClientCleanupToAdmin|AdminClientCloseFallsBackToClientWhenAdminIsNil|SyncProducerCloseClosesClientAndProducer|SyncProducerCloseStillClosesProducerWhenClientCloseFails)' -count=1`

No. The change only makes Kafka Sarama client cleanup more complete on close and on partially initialized failure paths.

No.

### Release note 

```release-note
Release Kafka Sarama clients on cleanup and initialization failures to avoid leaking background goroutines, broker connections, and related caches during retries.
```

### DIFF
--- a/pkg/sink/kafka/sarama_sync_producer.go
+++ b/pkg/sink/kafka/sarama_sync_producer.go
@@ -26,10 +26,21 @@ import (
 	"go.uber.org/zap"
 )
 
+type saramaSyncClient interface {
+	Brokers() []*sarama.Broker
+	Close() error
+}
+
+type saramaSyncProducerClient interface {
+	SendMessage(msg *sarama.ProducerMessage) (partition int32, offset int64, err error)
+	SendMessages(msgs []*sarama.ProducerMessage) error
+	Close() error
+}
+
 type saramaSyncProducer struct {
 	id       commonType.ChangeFeedID
-	client   sarama.Client
-	producer sarama.SyncProducer
+	client   saramaSyncClient
+	producer saramaSyncProducerClient
 	closed   *atomic.Bool
 }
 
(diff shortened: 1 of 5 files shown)


pkg/sink/kafka/sarama_sync_producer.go
```diff
@@ -26,10 +26,21 @@ import (
 	"go.uber.org/zap"
 )
 
+type saramaSyncClient interface {
```
3AceShowHand: redundant interface。

Issue #3413, opened 2025-12-01
Support `drain capture` in TiCDC New Arch


hongyunyan:
In the current TiCDC architecture, the drain capture command allows for the graceful migration of workloads (processors) from a specific node to others before maintenance or scaling down. This prevents synchronization spikes and service interruptions.

As we transition to the New Architecture, we need to ensure this parity to support seamless rolling upgrades and cluster maintenance.

**The drain capture functionality in the New Architecture should include**:

Node State Management: The ability to mark a Capture node as Draining. Once marked, the scheduler should stop assigning new workloads to this node.

Graceful Migration: Trigger the new scheduler to proactively move existing dispatchers (or tasks) from the draining node to other healthy nodes.

Status Tracking: Provide a way to monitor the progress (e.g., remaining task count on the node) via CLI or API.

Safe Shutdown: Ensure the node can be safely stopped once all workloads have been migrated (task count reaches zero).

PR #4522, opened 2026-03-18
coordinator,maintainer: add node liveness foundation

hongyunyan:

The drain-capture work in #4190 currently mixes node liveness, dispatcher drain runtime, and public API orchestration in one review unit. This PR extracts the node liveness foundation into the first stacked PR so the protocol and control loop can be reviewed independently.

Issue Number: ref https://github.com/pingcap/ticdc/issues/3413

Background:
- #4190 combines node liveness protocol, dispatcher evacuation, and the v1 drain API in a single change set.
- This split keeps the existing runtime design intact, but reduces review scope and keeps generated protobuf code aligned with the protocol change.

Motivation:
- Make the drain stack reviewable in smaller, mergeable layers.
- Land the liveness state machine before any scheduler or API depends on it.
- Keep election gating and coordinator retry logic in the same review unit as the liveness protocol.

Summary:
- add `NodeLiveness`, `NodeHeartbeat`, and `SetNodeLiveness*` protocol support
- make liveness transitions monotonic from `ALIVE` to `DRAINING` to `STOPPING`
- let coordinator track node liveness, observe epochs from heartbeat and response paths, and drive retry-based drain and stop progression
- gate server election when the current node is no longer `Alive`
- add focused unit coverage for the liveness controller, maintainer handling, and liveness state transitions

How it works:
- maintainers emit node liveness and epoch information in heartbeat and command response paths
- coordinator keeps an in-memory liveness view and retries transitions until the node can safely move forward
- scheduler drain behavior and the public drain API are intentionally left to follow-up stacked PRs

No new public API is introduced in this PR. The change only adds the node liveness foundation that later drain layers depend on, and keeps the protocol change paired with regenerated protobuf code.

No additional user-facing documentation is needed for this split. The decomposition follows the approved drain-capture split design and the existing drain design docs.

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  * Node drain management with monotonic liveness states (Alive → Draining → Stopping), heartbeat reporting, and coordinated transitions
  * Messaging/RPCs for node heartbeat and setting node liveness; coordinator advances drain progress and gates scheduling on liveness

* **Behavior Changes**
  * Maintainer and server flows now respect node liveness (skip/resign when not alive)
  * Health API reflects node liveness

* **Tests**
  * Added unit tests covering drain controller, maintainer liveness handling, and liveness state machine


coordinator/controller.go
```diff
@@ -262,6 +268,10 @@ func (c *Controller) onPeriodTask() {
 	for _, req := range requests {
 		_ = c.messageCenter.SendCommand(req)
 	}
+
+	c.drainController.AdvanceLiveness(func(id node.ID) bool {
+		return len(c.changefeedDB.GetByNodeID(id)) == 0 && c.operatorController.CountOperatorsInvolvingNode(id) == 0
```
wk989898: If `CountOperatorsInvolvingNode` just compares with zero, I think it's better to modify the function and return immediately when there exists an operator.

heartbeatpb/heartbeat.proto
```diff
@@ -149,6 +149,39 @@ message MaintainerStatus {
...
+    // DRAINING is excluded from destination selection and is being drained.
+    // In-flight operations may still complete, but no new leader campaign is allowed.
+    DRAINING = 1;
+    // STOPPING keeps destination exclusion and enforces shutdown safety.
+    // It rejects new maintainer creation, allows cleanup/remove
+    STOPPING = 2;
+}
+
+// NodeHeartbeat is sent periodically from a node to the coordinator.
+message NodeHeartbeat {
```
3AceShowHand: Shall we really introduce a new heartbeat? Can we make it reuse other heartbeat ?

3AceShowHand (approved):

3AceShowHand:
There is 2 questions:

1. what is the real difference beteween the DRAINING AND STOPPING, it looks really similar to me
2. Can we reuse exsting messages to carries liveness information ?

heartbeatpb/heartbeat.proto
```diff
@@ -149,6 +149,39 @@ message MaintainerStatus {
...
+// NodeLiveness is node-reported liveness.
+// Coordinator uses it for destination filtering and safe rolling-restart handoff
+// without persisting a separate intent in etcd.
+// It is monotonic and only allows: ALIVE -> DRAINING -> STOPPING.
+enum NodeLiveness {
+    // ALIVE accepts new destination placement, maintainer creation, and leader campaign/service.
+    ALIVE = 0;
+    // DRAINING is excluded from destination selection and is being drained.
+    // In-flight operations may still complete, but no new leader campaign is allowed.
+    DRAINING = 1;
```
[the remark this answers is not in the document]
hongyunyan: I assume Draining will reject new scheduling targets, but if the scheduling to move maintainer to the target  is beginning, we will accept that.
DRAINING means the node is in the drain phase: it should not receive new scheduling targets or start new leader campaigns, but in-flight work can still finish. STOPPING means the node has finished draining and is in the final shutdown phase, so it also rejects new maintainer creation and is considered safe to stop.

wk989898 (approved):

Issue #3943, opened 2026-01-07
Implement CMEK Encryption Support for TiCDC Next-Gen Architecture


tenfyzhong:
## Background

TiDB already supports Customer-Managed Encryption Keys (CMEK) encryption in the Next-Gen architecture. TiCDC needs to implement similar encryption capabilities to secure data before flushing it to disk during change data capture operations.

## Objectives

Implement CMEK encryption support for TiCDC Next-Gen with the following key components:

1. **Keyspace Encryption Metadata Retrieval**
   - Fetch keyspace encryption metadata from TiKV

2. **Master Key Management**
   - Retrieve master keys from the Key Management System (KMS)

3. **Data Encryption/Decryption**
   - Encrypt data using data-keys (EncryptionKey) before writing to disk
   - Decrypt data using the same data-key when reading from disk

## Design Overview

The implementation should follow a similar approach to TiKV's CMEK encryption design:

- When a changefeed is created, TiCDC fetches data from TiKV and writes it to disk
- Before writing, encrypt the data using the data-key (EncryptionKey)
- When reading, decrypt the data using the same data-key
- Use EncryptionMetaManager to manage data-keys with caching support
- If no cached key exists, retrieve it from TiKV and cache for future use

## Data Format Considerations

The implementation will add 4 bytes for encryption metadata, which may introduce backward incompatibility with previous implementations. This needs to be carefully handled to ensure smooth upgrades.

## Key Features

- Support for key rotation
- Efficient key caching mechanism
- Backward compatibility handling for existing deployments

PR #4573, opened 2026-03-23
cmek split 6/8: eventstore encrypted persistence

tenfyzhong:

Issue Number: ref #3943

This is split PR **6/8** from #3955.

This PR adds CMEK support to EventStore disk persistence:

- Wire optional encryption manager in EventStore.
- Encrypt persisted event payloads before writing pebble.
- Decrypt payloads during iterator reads.
- Propagate keyspace ID for encryption/decryption context.
- Add unit tests for encrypted EventStore path.

#### Tests

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

When CMEK is enabled, encryption/decryption introduces expected CPU overhead. Compatibility is preserved for unencrypted data paths.

No.

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  * Added optional end-to-end encryption support for events with optional compression integration.
  * Per-subscription encryption context support for enhanced configuration flexibility.
  * Maintained full backward compatibility with existing unencrypted event data.

* **Tests**
  * Extended test coverage for encryption functionality and related operations.


logservice/eventstore/event_store.go
```diff
@@ -1325,6 +1339,20 @@ func (e *eventStore) writeEvents(
 				metrics.EventStoreCompressedRowsCount.Inc()
 			}
 
+			// Encrypt if encryption is enabled (after compression)
+			if e.encryptionManager != nil {
+				encryptedValue, err := e.encryptionManager.EncryptData(context.Background(), event.keyspaceID, value)
```
lidezhu, line 1407: Is it possible to reuse buffer for encryption just like compression do?

tenfyzhong: It is possible, but not as a small change in `event_store.go`. Compression can reuse the caller-provided dst buffer, while the current encryption stack allocates internally in `pkg/encryption` for the IV, ciphertext, and header-encoded result.

If we want proper buffer reuse here, I'd prefer to add a dedicated reusable encryption buffer and an `EncryptDataTo`-style API in `pkg/encryption`, likely as a follow-up to keep this PR focused.

logservice/eventstore/event_store.go
```diff
@@ -1319,7 +1333,53 @@ func (e *eventStore) writeEvents(
 			valueBytesAfter := valueBytesBefore
 			keyLen := encodedKeyLen(kv)
 
-			if e.enableZstdCompression && valueBytesBefore > int64(e.compressionThreshold) {
+			if e.encryptionManager != nil {
```
lidezhu: This branch have a lot of duplicate code with `if e.enableZstdCompression && valueBytesBefore > int64(e.compressionThreshold)`. May be we can reorder the if statements and make `if e.encryptionManager == nil && !(e.enableZstdCompression && valueBytesBefore > int64(e.compressionThreshold))`(encryption disabled and compression disabled) as the first case?

logservice/eventstore/event_store.go
```diff
@@ -1429,6 +1520,15 @@ func (iter *eventStoreIter) Next() (*common.RawKVEntry, bool) {
 		key := iter.innerIter.Key()
 		value := iter.innerIter.Value()
 
+		// Decrypt if encrypted data is detected and encryption manager is available
+		if encryption.IsEncrypted(value) && iter.encryptionManager != nil {
```
lidezhu: `EncryptData` can legally return the original plaintext when encryption is disabled for the keyspace, or when `allow-degrade-on-error` is true and key lookup fails. However, `encryption.IsEncrypted(value)` only checks whether the first byte is non-zero. A normal plaintext `RawKVEntry` for a Put event starts with little-endian `OpType == 1`, so its first byte is also `0x01`. That means a plaintext event-store row written through the encryption-enabled path can be misclassified as encrypted during iteration, then passed into `DecryptData`, and finally panic on read.

logservice/eventstore/event_store.go
```diff
@@ -1279,6 +1293,124 @@ func (e *eventStore) collectAndReportStoreMetrics() {
 	metrics.EventStoreResolvedTsLagGauge.Set(eventStoreResolvedTsLagInSec)
 }
 
+type preparedValueForWrite struct {
+	value           []byte
+	compressionType CompressionType
+	valueBytesAfter int64
+	rawBuf          []byte
+	compressionBuf  []byte
+	directWrite     bool
```
lidezhu: It seems that with `directWrite`, `preparedValueForWrite` is not a good abstraction and instead makes the code harder to follow. I prefer the original structure. 
For example, for the most common case, we only need to understand a straightforward branch like:
```
if !compressed && !encrypted {
... 
} 
```
But with the current abstraction, we have to read both how `preparedValueForWrite` is produced and how it is consumed, including the special `directWrite` case where the prepared value is not actually used. That adds cognitive overhead without simplifying the write path.

logservice/eventstore/event_store.go
```diff
@@ -1435,6 +1452,60 @@ func (e *eventStore) writeEvents(
...
+		return rawValue, value, compressionType, nextRawBuf, nextDstBuf
+	}
+
+	maxEncodedSize := encoder.MaxEncodedSize(len(rawValue))
+	value = ensureValueBuffer(dstBuf, maxEncodedSize)
+	value = encoder.EncodeAll(rawValue, value)
+	compressionType = CompressionZSTD
+	nextDstBuf = value[:0]
+	metrics.EventStoreCompressedRowsCount.Inc()
+	return rawValue, value, compressionType, nextRawBuf, nextDstBuf
```
lidezhu: Seems the returned `rawValue` is never used?

lidezhu (approved):
LGTM with minor comments

3AceShowHand (approved):

3AceShowHand (approved):

flowbehappy (approved):

PR #4659, opened 2026-03-31
table route: add the table route core functionality

### LINKED ISSUE #4941: add table route core functionality
Implement the core table route functionality for TiCDC changefeeds:

- Match table route rules and map source schema/table names to target names.
- Apply routed names to table metadata and DDL events.
- Rewrite parser-supported DDL SQL so referenced table names stay consistent with routed metadata.
- Add tests based on real DDL events generated by `EventTestHelper`, including cross-database and multi-table DDL cases.
- Explicitly reject parser-unsupported index DDLs until parser-backed support is available.

3AceShowHand:

Issue Number: close #4941

This PR adds the core table route implementation used by changefeeds to map source schema/table names to target names and keep routed metadata consistent with routed DDL queries.

- Add `downstreamadapter/routing` with rule matching, schema/table placeholder substitution, case-sensitive matching, and ambiguity checks for schema-level DDL routing.
- Apply table route rules to `TableInfo`, `DDLEvent` primary/extra table fields, `MultipleTableInfos`, and blocked table names without mutating the original event metadata.
- Rewrite parser-supported DDL SQL through TiDB parser/AST visitors so all table references in the query are routed, including query-only references such as `CREATE VIEW ... SELECT ...` and `FOREIGN KEY ... REFERENCES ...`.
- Cover schema/table/view/partition/column/index/TTL/multi-schema/FK DDLs with real `EventTestHelper` generated DDL events, including cross-database references and multi-table rename variants.
- Reject CDC-local parser-unsupported DDL actions (`ADD FULLTEXT INDEX`, `CREATE HYBRID INDEX`) when table routing is enabled, so they fail explicitly until parser-backed support is added.

### Review Guide

Suggested review order:

1. `downstreamadapter/routing/router.go` — public routing behavior, when DDL parsing is triggered, and how routed events are constructed.
2. `downstreamadapter/routing/ddl_query_rewriter.go` — TiDB parser-backed table-name extraction and AST rewrite logic.
3. `pkg/common/event/ddl_event.go` and `pkg/common/event/util.go` — routed DDL event construction and test helper metadata generation.
4. `pkg/filter/ddl.go` — CDC-local unsupported DDL action handling.
5. Tests in `downstreamadapter/routing/*_test.go` — start with apply-level behavior, then DDL rewriter cases, then the supported-DDL matrix.

#### Tests 

- Unit test

Tested locally:

- `GOCACHE=/tmp/ticdc-go-build-cache go test --tags=intest ./downstreamadapter/routing -count=1`
- `GOCACHE=/tmp/ticdc-go-build-cache go test --tags='intest nextgen' ./downstreamadapter/routing -count=1`

No compatibility break is expected. When no table route rule is configured, DDL routing returns the original event without parser work. When table route is configured, DDL SQL is parsed so query-only table references can be routed correctly.

No user-facing documentation update in this PR. This PR adds the core implementation and test coverage.

### Release note 

```release-note
Add core table routing support for routed DDL event metadata and DDL query rewriting.
```


hongyunyan (approved):

downstreamadapter/eventcollector/dispatcher_stat.go
```diff
@@ -538,6 +538,18 @@ func (d *dispatcherStat) handleSingleDataEvents(events []dispatcher.DispatcherEv
 	}
 	if events[0].GetType() == commonEvent.TypeDDLEvent {
 		ddl := events[0].Event.(*commonEvent.DDLEvent)
+		ddl, err := d.target.GetRouter().ApplyToDDLEvent(ddl)
```
wk989898, line 541: In some cases, is the router nil?

downstreamadapter/routing/ddl_query_rewriter.go
```diff
@@ -0,0 +1,250 @@
...
+)
+
+// rewriteParserBackedDDLQuery rewrites a parser-supported DDL query by applying routing rules.
+func (r Router) rewriteParserBackedDDLQuery(ddl *commonEvent.DDLEvent) (string, error) {
+	if len(r.rules) == 0 {
+		return ddl.Query, nil
+	}
+
+	queries := []string{ddl.Query}
+	if strings.Contains(ddl.Query, ";") {
```
wk989898: It is wrong to split multiple queries by a semicolon. Because the comment may contain semicolons. `CREATE TABLE t (
  c VARCHAR(100) DEFAULT 'a;b;c'
);`

downstreamadapter/eventcollector/dispatcher_stat.go
3AceShowHand: router will not be nil, it's not a pointer, but instance.

wk989898 (approved):

flowbehappy (approved):

PR #4663, opened 2026-03-31
downstreamadapter,config,api: wire event collector batch config

### LINKED ISSUE #4777: event collector support config the dynamic stream batch
as title

3AceShowHand:

Issue Number: close #4777 

This PR wires event collector batch config through the API, config, dispatcher manager, and event collector paths.

- add `event_collector_batch_count` and `event_collector_batch_bytes` to the v2 replica config and internal replica config
- preserve `unset` versus explicit `0` by keeping these fields as `*int` in runtime `ChangefeedConfig`
- resolve batch count and batch bytes from sink defaults in dispatcher manager, and apply explicit changefeed overrides when configured
- pass the resolved batch config through dispatcher shared state into event collector dynstream area settings
- add tests for API conversion, config propagation, validation, precedence, and event collector batching behavior

#### Tests

- Unit test
  - `go test ./api/v2 -run 'TestReplicaConfigConversionBatchFields$'`
  - `go test ./pkg/config -run 'Test(ChangeFeedInfoToChangefeedConfigBatchFields|ReplicaConfigValidateBatchConfig)$'`
  - `go test ./downstreamadapter/dispatchermanager -run 'TestDispatcherManagerBatchConfig$'`

No compatibility change for existing users when these fields are unset, because the default path still uses sink derived settings. When configured, the behavior changes only for the target changefeed.

No.

### Release note

```release-note
None
```

## Release Notes

* **New Features**
  * Added configurable event collector batching with optional parameters for batch count and batch bytes limits.
  * Enables fine-tuning of event collection behavior through replica and changefeed configuration.

* **Tests**
  * Added comprehensive test coverage for batch configuration validation, conversion, and application.


asddongmen (approved):

pkg/config/changefeed.go
```diff
@@ -188,9 +188,11 @@ type ChangefeedConfig struct {
 	TimeZone      string `json:"timezone" default:"system"`
 	CaseSensitive bool   `json:"case_sensitive" default:"false"`
 	// if true, force to replicate some ineligible tables
-	ForceReplicate bool          `json:"force_replicate" default:"false"`
-	Filter         *FilterConfig `toml:"filter" json:"filter"`
-	MemoryQuota    uint64        `toml:"memory-quota" json:"memory-quota"`
+	ForceReplicate           bool          `json:"force_replicate" default:"false"`
+	Filter                   *FilterConfig `toml:"filter" json:"filter"`
```
wk989898, line 192: Why does this field need `toml`?

pkg/config/changefeed.go
```diff
@@ -188,9 +188,11 @@ type ChangefeedConfig struct {
...
+	MemoryQuota              uint64        `toml:"memory-quota" json:"memory-quota"`
+	EventCollectorBatchCount *int          `json:"event_collector_batch_count"`
```
wk989898, line 194: Does it need the default value?

3AceShowHand: No，level it to be nil, and then it can be derived from the sink configuration in the runtime.

wk989898 (approved):

flowbehappy (approved):

PR #4674, opened 2026-04-01
dynstream: support area batch config and count+bytes batching

### LINKED ISSUE #4675: dynamic stream support batch by bytes
as title

3AceShowHand:

Issue Number: close #4675 

This is the 2rd PR to support dynamic stream batch events by bytes at the area level.

dynstream currently only uses a shared batch config and does not have a common `count + bytes` batching core.

- add area-level batch config support in dynstream
- add the core batcher for `count + bytes` batching
- make event queue and stream use the new batching logic
- add unit tests for batch config and batching behavior

#### Tests

- Unit test

No obvious compatibility impact. This PR only changes dynstream internal batching logic.

No.

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  * Per-area batch configuration with independent event-count and byte-size limits; streams and queues apply area-specific batching.

* **API Changes**
  * New area settings constructor to specify batch limits; handler callback return type and area-settings parameter names clarified.
  * Batch semantics: non-positive counts default to 1; negative byte limits treated as 0.

* **Tests**
  * Added and updated unit tests covering per-area batching, byte/count boundaries, first-add semantics, and edge cases.

### DIFF
--- a/utils/dynstream/area_batch_config_registry.go
+++ b/utils/dynstream/area_batch_config_registry.go
@@ -1,4 +1,4 @@
-// Copyright 2025 PingCAP, Inc.
+// Copyright 2026 PingCAP, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


utils/dynstream/area_batch_config_registry.go
```diff
@@ -0,0 +1,110 @@
...
+	cfg = newBatchConfig(cfg.softCount, cfg.hardBytes)
+	if cfg == s.defaultConfig {
+		return
+	}
+
+	s.setOverrideLocked(area, cfg)
+}
+
+// onRemovePath is not thread safe.
+func (s *areaBatchConfigRegistry[A]) onRemovePath(area A) {
```
wk989898: ditto

3AceShowHand: the lock is called by the outside caller.

PR #4698, opened 2026-04-03
logpuller: optimize resolved-ts dispatch allocations

### LINKED ISSUE #4697: logpuller: excessive CPU usage under resolved-ts-heavy workload due to fixed-size slice allocations
### Problem

  In `dispatchResolvedTsEvent` (`logservice/logpuller/region_request_worker.go`), we always allocate a
  `[]*regionFeedState` with capacity `1024`, and allocate another same-size slice after each `flush`.

  This is inefficient when resolved-ts batches are tiny (most often 1 region), and can cause high short-lived
  allocation pressure.

  ### Impact

  Under resolved-ts-heavy workloads, CPU is dominated by GC/runtime overhead instead of business logic.

  Observed symptoms:
  - High `process_cpu_seconds_total`
  - GC-heavy CPU profile (`runtime.scanobject`, `runtime.findObject`, etc.)
  - Allocation hotspot in `dispatchResolvedTsEvent`

Image

Image

Image

asddongmen:

Issue Number: close #4697 

  - Added a **single-region fast path**:
    - If `len(resolvedTsEvent.Regions) == 1`, directly push one-state event.
  - Changed initial `resolvedStates` capacity to:
    - `capHint = min(len(resolvedTsEvent.Regions), resolvedTsStateBatchSize)`
  - Kept large-batch protection:
    - `resolvedTsStateBatchSize` remains `1024`.
  - Removed redundant allocation pattern:
    - After `flush`, do not always allocate a new slice.
    - Allocate the next batch buffer only when there are remaining regions to process.
  - Added tests and benchmarks in `region_request_worker_test.go`:
    - Single-region fast-path behavior
    - Large-batch split behavior
    - Legacy vs current benchmark comparisons

**Benchmark Command**

```bash
mkdir -p /tmp/go-cache /tmp/go-tmp && \
GOCACHE=/tmp/go-cache GOTMPDIR=/tmp/go-tmp \
go test ./logservice/logpuller \
  -run '^$' \
  -bench 'BenchmarkDispatchResolvedTsEvent' \
  -benchmem \
  -count=5
```

**Average Comparison (5 runs)**

| Scenario | Legacy ns/op | Current ns/op | ns/op Change | Speedup | Legacy B/op | Current B/op | B/op Change | Legacy allocs/op | Current allocs/op | allocs Change |
|:--|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| SingleRegion | 664.94 | 30.19 | -95.46% | 22.02x | 19074 | 137 | -99.28% | 2.00 | 1.00 | -50.00% |
| SmallBatch | 658.94 | 148.20 | -77.51% | 4.45x | 19071 | 256 | -98.66% | 2.00 | 1.00 | -50.00% |
| LargeBatch | 36325.80 | 36532.80 | +0.57% | 0.99x | 47899 | 38424 | -19.78% | 5.00 | 4.00 | -20.00% |

- `SingleRegion` and `SmallBatch` show major CPU and allocation improvements.
- `LargeBatch` keeps bounded memory behavior and reduces allocations; `ns/op` is roughly flat (+0.57%).

#### Tests 

- Unit test

### Release note 

```release-note
  Reduce CPU and allocation overhead in log puller by optimizing resolved-ts event dispatch for small batches.
```

## Summary by CodeRabbit

* **Performance**
  * Improved resolved-timestamp dispatch: smarter pre-sizing, buffer reuse, and adaptive batching for single-region fast path and large multi-region splits, reducing allocations and improving throughput.

* **Tests**
  * Added unit tests and benchmarks validating single-region fast-path, large-set batch-splitting, and performance comparisons against the legacy dispatcher.

### DIFF
--- a/logservice/logpuller/region_request_worker.go
+++ b/logservice/logpuller/region_request_worker.go
@@ -301,23 +301,6 @@ func (s *regionRequestWorker) dispatchResolvedTsEvent(resolvedTsEvent *cdcpb.Res
 	// Avoid allocating a huge states slice when resolvedTsEvent.Regions is large.
 	// Push resolved-ts events in batches to reduce peak memory usage and improve GC behavior.
 	const resolvedTsStateBatchSize = 1024
-	if len(resolvedTsEvent.Regions) == 1 {
-		regionID := resolvedTsEvent.Regions[0]
-		if state := s.getRegionState(subscriptionID, regionID); state != nil {
-			s.client.pushRegionEventToDS(subscriptionID, regionEvent{
-				resolvedTs: resolvedTsEvent.Ts,
-				states:     []*regionFeedState{state},
-			})
-			return
-		}
-		log.Warn("region request worker receives a resolved ts event for an untracked region",
-			zap.Uint64("workerID", s.workerID),
-			zap.Uint64("subscriptionID", uint64(subscriptionID)),
-			zap.Uint64("regionID", regionID),
-			zap.Uint64("resolvedTs", resolvedTsEvent.Ts))
-		return
-	}
-
 	capHint := len(resolvedTsEvent.Regions)
 	if capHint > resolvedTsStateBatchSize {
 		capHint = resolvedTsStateBatchSize
--- a/logservice/logpuller/region_request_worker.go
+++ b/logservice/logpuller/region_request_worker.go
@@ -339,7 +339,11 @@ func (s *regionRequestWorker) dispatchResolvedTsEvent(resolvedTsEvent *cdcpb.Res
 			if len(resolvedStates) >= resolvedTsStateBatchSize {
 				flush()
 				if i+1 < len(resolvedTsEvent.Regions) {
-					resolvedStates = make([]*regionFeedState, 0, resolvedTsStateBatchSize)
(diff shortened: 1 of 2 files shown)


logservice/logpuller/region_request_worker.go
```diff
@@ -301,23 +301,50 @@ func (s *regionRequestWorker) dispatchResolvedTsEvent(resolvedTsEvent *cdcpb.Res
 	// Avoid allocating a huge states slice when resolvedTsEvent.Regions is large.
 	// Push resolved-ts events in batches to reduce peak memory usage and improve GC behavior.
 	const resolvedTsStateBatchSize = 1024
-	resolvedStates := make([]*regionFeedState, 0, resolvedTsStateBatchSize)
+	if len(resolvedTsEvent.Regions) == 1 {
```
lidezhu: It seems we don't need to treat len(resolvedTsEvent.Regions) == 1 as a special case, as the later logic should handle it just fine.

hongyunyan (approved):

lidezhu (approved):

PR #4759, opened 2026-04-07
maintainer,heartbeatpb: add drain target plumbing

hongyunyan:

The dispatcher-drain work split out of #4190 still mixes the drain target protocol, maintainer runtime behavior, coordinator scheduling, and public API orchestration in one review path. This PR extracts the drain target protocol and maintainer-manager plumbing into a dedicated review unit so reviewers can focus on how the target is propagated and acknowledged before looking at scheduler behavior.

Issue Number: ref https://github.com/pingcap/ticdc/issues/3413

Background:
- #4522 already merged the node liveness foundation into `master`.
- The remaining dispatcher-drain work from the old `#4523` still mixed protocol, runtime scheduling, and API orchestration.
- This PR starts a new stacked split from `master` and keeps only the drain target plumbing.

Motivation:
- Separate protocol and manager-level state propagation from drain scheduling logic.
- Let reviewers verify target monotonicity and acknowledgment rules before reading scheduler algorithms.
- Keep protobuf and messaging changes in the same review unit as the maintainer-manager handlers that consume them.

Summary:
- add `SetDispatcherDrainTargetRequest`
- extend node heartbeat with dispatcher-drain target acknowledgment fields
- add manager-level drain target state with monotonic update rules and clear-once behavior
- fan out the latest drain target from maintainer manager to newly created and already running maintainers
- add focused tests for target apply, stale-update rejection, and heartbeat acknowledgment

How it works:
- coordinator can send a dispatcher-drain target to the maintainer manager
- the maintainer manager applies it at node scope and reports the applied target back in node heartbeat
- the latest target is also seeded into each local maintainer so later runtime layers can consume a consistent snapshot

#### Tests

- Unit test
  - `go test ./maintainer -run 'TestSetDispatcherDrainTarget|TestSetNodeLiveness'`
  - `go test ./pkg/messaging ./heartbeatpb`

This PR does not add a user-facing API by itself. It only adds the internal protocol and manager-level drain target propagation needed by the later runtime and API layers.

No additional user-facing documentation is needed for this split. It is an internal decomposition of the drain-capture implementation.

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  * Coordinators can set or clear dispatcher drain targets (node + epoch); nodes acknowledge these targets in heartbeats and report per-target drain progress.

* **Improvements**
  * Scheduling avoids evacuated targets and pauses/extends rebalancing during active drain windows to reduce disruption.

* **Tests**
  * Added tests for drain-target apply/clear/ordering, heartbeat acknowledgements, and maintainer validation behavior.


maintainer/maintainer_manager_maintainers.go
```diff
@@ -0,0 +1,338 @@
...
+			return &heartbeatpb.MaintainerStatus{
+				ChangefeedID: req.GetId(),
+				State:        heartbeatpb.ComponentState_Stopped,
+			}
+		}
+
+		// It's cascade remove, we should remove the dispatcher from all node.
+		// Here we create a maintainer to run the remove dispatcher logic.
+		maintainer = NewMaintainerForRemove(changefeedID, p.conf, p.nodeInfo, p.taskScheduler, req.KeyspaceId)
+		p.registry.Store(changefeedID, maintainer)
```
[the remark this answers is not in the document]
hongyunyan: This write path is serialized today, so I do not think the `Store` here can race with another cascade-remove `Store` in the current design.

All add/remove control messages first go through `recvMessages()` into `m.msgCh`, and `Manager.Run()` processes `handleMessage()` from that channel on a single goroutine. That means `handleRemoveMaintainer()` is not executed concurrently for two control messages of the same changefeed. The other concurrent paths touching `registry` (`dispatchMaintainerMessage`, heartbeat building, cleanup iteration) only `Load`/`Range`; they do not create another remove maintainer for the same key.

So while `LoadOrStore` would be a reasonable defensive hardening if we ever make the write path concurrent, I do not see a real goroutine leak/overwrite race here with the current manager event-loop model.

maintainer/maintainer_manager.go
```diff
@@ -178,13 +160,9 @@ func (m *Manager) sendMessages(msg *heartbeatpb.MaintainerHeartbeat) {
 	}
 }
 
-// Close closes, it's a block call
+// Close closes, it's a block call.
```
3AceShowHand: this comment is funny, Close closes.

maintainer/maintainer_manager_maintainers.go
```diff
@@ -0,0 +1,338 @@
...
+// maintainers, routes maintainer-bound messages, and aggregates maintainer
+// heartbeats back to coordinator.
+//
+// In contrast, managerNodeState owns node-scoped state shared by the whole
+// capture, such as liveness, node epoch, and the latest manager-level drain
+// target. The Manager combines both layers: managerNodeState is the single
+// node-wide source of truth, while managerMaintainerSet fans that node-scoped
+// state out to individual maintainers and manages their per-changefeed
+// lifecycles.
+type managerMaintainerSet struct {
```
3AceShowHand, line 45: maintainer_manager_maintainer.go, so weird.

maintainer/maintainer_manager_maintainers.go
```diff
@@ -0,0 +1,338 @@
...
+	// During liveness propagation and scheduler view convergence, in-flight operators can still
+	// issue add requests to this node; rejecting them here may break drain convergence.
+	// Only STOPPING performs a hard reject for new maintainer creation.
+	currentLiveness := liveness.CaptureAlive
+	if m.node.liveness != nil {
+		currentLiveness = m.node.liveness.Load()
+	}
+	if currentLiveness == liveness.CaptureStopping {
+		changefeedID := common.NewChangefeedIDFromPB(req.Id)
+		log.Info("reject add maintainer request because node is stopping",
```
3AceShowHand, line 78: make sure this is not verbose

maintainer/maintainer_manager_node.go
```diff
@@ -117,11 +153,76 @@ func (m *Manager) onSetNodeLivenessRequest(msg *messaging.TargetMessage) {
 	m.sendSetNodeLivenessResponse(current)
 }
 
+// onSetDispatcherDrainTargetRequest updates the latest dispatcher drain target
+// and forwards it to all existing maintainers. A manager-level node heartbeat
+// is sent after each accepted update so coordinator can observe the ack even
+// when this node currently hosts no maintainers.
+func (m *Manager) onSetDispatcherDrainTargetRequest(msg *messaging.TargetMessage) {
+	if m.coordinatorID != msg.From {
+		log.Warn("ignore set dispatcher drain target request from non coordinator",
```
3AceShowHand, line 162: will this happen? If this happen, will it be verbose? Is this helpful to detect problems?

3AceShowHand (approved):

maintainer/scheduler/drain_common.go
```diff
@@ -0,0 +1,183 @@
...
+// allocation on scheduler hot paths.
+func filterNodeIDsByDrainTarget(
+	nodeIDs []node.ID,
+	state drainStateSnapshot,
+) []node.ID {
+	target, _, active := activeDrainTarget(state)
+	if !active {
+		return nodeIDs
+	}
+	filtered := nodeIDs[:0]
```
wk989898, line 154: This should clear the element with `nodeIDs`, is that expected?

maintainer/maintainer_manager.go
```diff
@@ -139,22 +134,9 @@ func (m *Manager) Run(ctx context.Context) error {
...
-					cf.Close()
-					log.Info("maintainer removed, remove it from dynamic stream",
-						zap.Stringer("changefeedID", cf.changefeedID),
-						zap.Uint64("checkpointTs", cf.getWatermark().CheckpointTs),
-					)
-					m.maintainers.Delete(key)
-				}
-				return true
-			})
+			m.sendHeartbeat()
```
wk989898, line 138: Why do we need to adjust the order of sendNodeHeartbeat and sendHeartbeat?

maintainer/maintainer_manager_maintainers.go
hongyunyan: I kept the current split in this PR. This file owns the changefeed-scoped half of manager state (registry, add/remove, maintainer heartbeat fanout), while `maintainer_manager_node.go` owns the node-scoped half. Renaming it here would be churn-only in a stacked PR and would not change the boundary itself.

maintainer/maintainer_manager.go
hongyunyan: The intent here is to prioritize the manager-level ack path. `sendNodeHeartbeat` is the authoritative channel for the node-scoped drain-target snapshot, while `sendHeartbeat` is the per-changefeed status fanout built on top of that node-scoped state. Sending the node heartbeat first lets coordinator observe the top-level snapshot as early as possible on each tick, even when there are no maintainers or when maintainer heartbeat aggregation is comparatively heavier. So there is no hidden dependency between the two calls, but the order is intentional to keep the control-plane ack path ahead of the lower-level status path.

maintainer/maintainer_manager_node.go
hongyunyan: Yes, this can happen on stale or misrouted control messages during coordinator changeover, and in that case I want a warning. This request is coordinator-only, so receiving it from another sender means a control-plane invariant was violated. It is not on a hot periodic path either: we only send this request when the drain target changes, so the warning should stay low-volume and useful for diagnosis. This is also consistent with the existing `onSetNodeLivenessRequest` guard.

maintainer/scheduler/drain_common.go
hongyunyan: Yes, that is expected. `filtered := nodeIDs[:0]` is doing in-place compaction: it reuses the input backing array and overwrites the visible prefix with the kept node IDs. Callers must use the returned slice, not the original length. In this scheduler helper the candidate slice is ephemeral, so reusing the storage is intentional to avoid an extra allocation on a hot path.

wk989898 (approved):

PR #4808, opened 2026-04-13
maintainer: fast retry temporarily ignored WAITING statuses

### LINKED ISSUE #4810: `CREATE TABLE ... LIKE ...` may incur an extra ~5s wait before the barrier advances
### What did you do?

- Execute a `CREATE TABLE ... LIKE ...` DDL.
- On the code path, this kind of DDL is recorded by schema store as a barrier event involving both the table-trigger dispatcher and the referenced-table dispatcher, rather than a single-dispatcher DDL.

- Once the dispatchers participating in the same barrier have received the DDL, the system should move to `Action_Write` or `Action_Pass` within normal scheduling cycles.
- The table-trigger dispatcher should not show a stable multi-second idle wait after the DDL has already been delivered.
- `PostFlush()` should run promptly, then the path should be woken up and later events on the same path should continue.

### What did you see instead?

- The table-trigger dispatcher receives the DDL quickly, but it attaches `wakeCallback` to `PostFlush()` and waits on the blocking-DDL path for a barrier action, so merely receiving the DDL does not unblock the path.
- For `CREATE TABLE ... LIKE ...`, the referenced-table dispatcher also receives the same DDL and reports it to the maintainer as a block event.
- The observed idle gap from the referenced-table dispatcher's first receipt of the DDL to barrier writer selection is about `5.011s`.
- During this interval, the barrier remains in `selected=false`, so no `write/pass` action is emitted; the table-trigger path therefore stays blocked and later events on the same path keep queueing.
- From the user perspective, the DDL has already been sent by the event broker and received by dispatchers, but end-to-end completion still shows an extra, near-fixed plateau of about 5 seconds.

### Versions of the cluster

```console
master
```

zier-one:

Issue Number: close #4810

When a non-DDL dispatcher reports a `WAITING` block status while the maintainer still sees that task as temporarily non-replicating, the status used to be silently ignored. The dispatcher then had to wait for the regular 5-second resend loop before retrying, which unnecessarily delayed barrier recovery and convergence.

1. Add `IgnoredBlockStatus` to `heartbeatpb.DispatcherStatus` so the maintainer can explicitly tell a dispatcher that its current `WAITING` status is temporarily ignored and should be retried soon, and regenerate `heartbeat.pb.go`.
2. On the dispatcher side, `HandleDispatcherStatus` now has an ignored-block branch. It schedules a fast retry only when the hint strictly matches the current in-flight `WAITING` block event.
3. `ResendTask` keeps the original 5-second periodic resend as the fallback path and adds a one-shot fast retry starting at 50ms with exponential backoff up to the same interval; ACK still cancels both the slow and fast resend paths.

Before:
img_v3_0210n_60180ab7-f08c-4b1a-8949-d3b32cf6249g

After:
img_v3_0210o_181a85fa-6f50-425f-80a7-b6545799a4ag

#### Tests 

[x] Unit test

### Release note 

```release-note
Fix barrier retry latency by triggering a fast retry when the maintainer temporarily ignores a dispatcher's `WAITING` block status instead of waiting for the regular resend interval.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents spurious blocked events by temporarily treating certain block-status reports from non-replicating dispatchers as ignored and preserving span state.

* **New Features**
  * Heartbeat messages now include an IgnoredBlockStatus hint and influenced-dispatcher info so receivers can fast-retry pending block events.

* **Tests**
  * Added tests for ignored-block handling, fast-resend timing, retry cancellation on ACK, and status propagation.


wk989898:
If there are lots of barrier events, does this change cause OOM?

maintainer/barrier.go
```diff
@@ -238,6 +240,195 @@ func (b *Barrier) Resend() []*messaging.TargetMessage {
...
+			log.Info("drop deferred block status because dispatcher moved to another node",
+				zap.String("changefeed", entry.value.cfID.Name()),
+				zap.String("dispatcher", entry.dispatcherID.String()),
+				zap.String("from", entry.value.from.String()),
+				zap.String("taskNode", task.GetNodeID().String()),
+				zap.Uint64("commitTs", entry.key.blockTs),
+				zap.Int64("mode", b.mode))
+			continue
+		}
+		if dispatcherAlreadyPassedPendingState(task, entry.value.state) {
```
zier-one: There is cache cleanup logic here. Generally speaking, there is at most one cache entry per dispatcher, so I understand that the memory usage is relatively controllable.

zier-one:
> If there are lots of barrier events, does this change cause OOM?

https://github.com/pingcap/ticdc/pull/4808#discussion_r3073474177

maintainer/barrier.go
```diff
@@ -238,6 +240,195 @@ func (b *Barrier) Resend() []*messaging.TargetMessage {
...
+				zap.String("changefeed", entry.value.cfID.Name()),
+				zap.String("dispatcher", entry.dispatcherID.String()),
+				zap.Uint64("commitTs", entry.key.blockTs),
+				zap.Duration("deferredDuration", time.Since(entry.value.firstSeenAt)),
+				zap.Int64("mode", b.mode))
+			continue
+		}
+
+		if needACK {
+			if _, ok := acks[entry.value.from]; !ok {
```
wk989898: `entry.value.from` can be extracted as a param

wk989898 (approved):

maintainer/barrier.go
```diff
@@ -83,11 +83,29 @@ func (b *Barrier) HandleStatus(from node.ID,
...
 				continue
 			} else {
 				if !b.spanController.IsReplicating(task) {
-					log.Info("Get block status from unreplicating dispatcher, ignore it", zap.String("changefeed", request.ChangefeedID.GetName()), zap.String("dispatcher", dispatcherID.String()), zap.Uint64("commitTs", status.State.BlockTs), zap.Int64("mode", b.mode))
+					log.Info("Get block status from unreplicating dispatcher, ignore it",
+						zap.String("changefeed", request.ChangefeedID.GetName()),
+						zap.String("dispatcher", dispatcherID.String()),
+						zap.Uint64("commitTs", status.State.BlockTs),
+						zap.Int64("mode", b.mode))
+					dispatcherStatus = append(dispatcherStatus, &heartbeatpb.DispatcherStatus{
```
hongyunyan, line 104: Please add a comment here to explain why we need ignoredBlockStatus

downstreamadapter/dispatcher/helper.go
```diff
@@ -282,10 +309,60 @@ func (t *ResendTask) Execute() time.Time {
 	return time.Now().Add(resendTimeInterval)
 }
 
+func (t *ResendTask) ScheduleFastRetry() {
```
hongyunyan: As I understand it, there is no need to implement exponential backoff here. Since a retransmission is triggered only upon receiving an "ignore" signal, the absence of exponential backoff will not result in a flood of retransmissions; therefore, we can employ a simpler retransmission logic.

downstreamadapter/dispatcher/helper.go
```diff
@@ -245,17 +262,27 @@ type HeartBeatInfo struct {
...
 	message      *heartbeatpb.TableSpanBlockStatus
 	dispatcher   Dispatcher
 	callback     func() // function need to be called when the task is cancelled
 	taskHandle   *threadpool.TaskHandle
 	executeCount uint64
+
+	mu               sync.Mutex
+	fastTaskHandle   *threadpool.TaskHandle
+	fastRetryAttempt uint32
+	closed           bool
```
hongyunyan: Why do we need closed field here?

downstreamadapter/dispatcher/basic_dispatcher.go
```diff
@@ -765,7 +766,19 @@ func (d *BasicDispatcher) HandleDispatcherStatus(dispatcherStatus *heartbeatpb.D
 		d.cancelResendTask(identifier)
 	}
 
-	// Step2: deal with the dispatcher action
+	// Step2: deal with the ignored block status
+	ignoredBlockStatus := dispatcherStatus.GetIgnoredBlockStatus()
```
hongyunyan, line 773: If it has ignoredBlockStatus, it should not have ack and action, so we could just return early?

downstreamadapter/dispatcher/helper.go
```diff
@@ -269,6 +292,23 @@ func newResendTask(message *heartbeatpb.TableSpanBlockStatus, dispatcher Dispatc
 }
 
 func (t *ResendTask) Execute() time.Time {
+	t.execute()
+	return time.Now().Add(resendTimeInterval)
+}
+
+func (t *ResendTask) ScheduleFastRetry() {
```
hongyunyan: Could we just call execute directly? why do we use a new handle?

hongyunyan (approved):

PR #4815, opened 2026-04-13
downstreamadapter: preserve remove upgrade during close

### LINKED ISSUE #4825: downstreamadapter: removed=true close upgrade can be lost during request de-duplication
## Background

`DispatcherOrchestrator` de-duplicates maintainer requests by `(changefeedID, messageType)` to avoid repeated retries flooding the queue.

For `MaintainerCloseRequest`, `removed=true` carries stronger semantics than `removed=false`, because it requires remove-only cleanup in addition to the normal close path.

## Problem

Before PR #4815, a later `removed=true` close request could still be lost if an earlier `removed=false` request for the same key had already been popped for processing.

The old queue semantics kept the key in the pending set while the worker was processing it:

- `Pop()` returned only the key
- the message itself still lived in `pending[key]`
- `Done(key)` deleted the pending entry after processing finished

That allowed this failure mode:

1. a `removed=false` close request is queued
2. the worker pops it and starts processing it
3. a later `removed=true` close request arrives for the same `(changefeedID, messageType)` key
4. the pending slot is upgraded to the stronger request
5. when the in-flight `removed=false` request finishes, `Done(key)` deletes the entire pending entry
6. the upgraded `removed=true` request is dropped without being re-queued
7. downstream only executes the normal close flow, while remove-only cleanup is skipped

## Impact

The stronger remove semantics can be silently lost.

That means a late remove upgrade may fail to trigger cleanup paths that are only required when the changefeed is actually removed.

## Expected behavior

If a later `removed=true` request upgrades an in-flight or queued `removed=false` close request, the stronger semantics should still be preserved and processed in a later round.

## Reference

- Fixed by PR #4815

hongyunyan:

Issue Number: close https://github.com/pingcap/ticdc/issues/4825

## Background
The dispatcher orchestrator de-duplicates close requests by changefeed and message type. A later `removed=true` close request could overwrite the pending entry for the same key while the earlier `removed=false` request was already in flight.

## Motivation
That overwrite was not re-queued, and `Done(key)` later deleted the upgraded request together with the in-flight one. As a result, the stronger remove semantics could be dropped silently and the downstream cleanup path would only execute the normal close flow.

## Summary of changes
- split the pending close queue state into `queued` and `inFlight` slots so an upgrade never overwrites the request a worker is already processing
- re-queue upgraded `removed=true` close requests for the next round and keep the in-flight request stable until `Done`
- make `DispatcherManager` keep `removed=true` as a sticky close requirement and finish remove-only cleanup even after the base close path has completed
- add a retryable MySQL remove cleanup helper so ddl_ts cleanup still works after the normal sink close releases the long-lived DB handle
- add regression tests for both the queue upgrade timing and the post-close remove cleanup path

## Validation
- `make fmt`
- `go test ./downstreamadapter/dispatcherorchestrator`
- `go test ./downstreamadapter/dispatchermanager`
- `go test --tags=intest ./downstreamadapter/sink/mysql`

## Release Notes

**Bug Fixes**
- Enhanced changefeed removal and cleanup processes to properly handle late removal requests and ensure cleanup completes even after initial shutdown
- Improved message queue dequeue operation to return message content directly, enabling more reliable request processing and ordering
- Refined message queue behavior for more predictable duplicate prevention and request ordering during queue operations

### Release note 

```release-note
None

### DIFF
--- a/downstreamadapter/dispatchermanager/dispatcher_manager.go
+++ b/downstreamadapter/dispatchermanager/dispatcher_manager.go
@@ -848,10 +848,7 @@ func (e *DispatcherManager) TryClose(removeChangefeed bool) bool {
 		e.removeChangefeedRequested.Store(true)
 	}
 	if e.closed.Load() {
-		if !e.cleanupRemovedChangefeed() {
-			return false
-		}
-		return e.isCloseComplete()
+		return e.cleanupRemovedChangefeed()
 	}
 	if e.closing.Load() {
 		return false
@@ -937,10 +934,6 @@ func (e *DispatcherManager) close() {
 	e.closed.Store(true)
 	log.Info("event dispatcher manager closed",
 		zap.Stringer("changefeedID", e.changefeedID))
-}
-
-func (e *DispatcherManager) isCloseComplete() bool {
-	return e.closed.Load() && (!e.removeChangefeedRequested.Load() || e.removeChangefeedCleaned.Load())
 }
 
 func (e *DispatcherManager) cleanupRemovedChangefeed() bool {
--- a/downstreamadapter/dispatchermanager/dispatcher_manager.go
+++ b/downstreamadapter/dispatchermanager/dispatcher_manager.go
@@ -139,8 +139,11 @@ type DispatcherManager struct {
 	// removeChangefeedCleaned records whether the remove-only cleanup has finished.
 	// closed=true only means the base close path completed; remove cleanup may still be pending.
 	removeChangefeedCleaned atomic.Bool
-	cancel                  context.CancelFunc
-	wg                      sync.WaitGroup
+	// removeChangefeedCleanupRunning prevents duplicate background cleanup runs
+	// while retries keep asking for remove semantics after the base close path ends.
+	removeChangefeedCleanupRunning atomic.Bool
+	cancel                         context.CancelFunc
+	wg                             sync.WaitGroup
 
 	// removeTaskHandles stores the task handles for async dispatcher removal
 	// map[common.DispatcherID]*threadpool.TaskHandle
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

File: downstreamadapter/dispatchermanager/dispatcher_manager.go [hunk 1]
```diff
@@ -902,11 +917,9 @@ func (e *DispatcherManager) close(removeChangefeed bool) {
 	log.Info("shared info closed", zap.Stringer("changefeedID", e.changefeedID))
 
 	if e.IsRedoEnabled() {
-		e.redoSink.Close(removeChangefeed)
-		// FIXME: cleanup redo log when remove the changefeed
-		e.closeRedoMeta(removeChangefeed)
```
wk989898, line 907: Why remove this?

downstreamadapter/dispatcherorchestrator/dispatcher_orchestrator.go
```diff
@@ -119,12 +119,11 @@ func getPendingMessageKey(msg *messaging.TargetMessage) (pendingMessageKey, bool
 // handleMessages processes messages from the queue
 func (m *DispatcherOrchestrator) handleMessages() {
 	for {
-		key, ok := m.msgQueue.Pop()
+		_, msg, ok := m.msgQueue.Pop()
```
wk989898:
```suggestion
		msg, ok := m.msgQueue.Pop()
```

wk989898 (approved):

downstreamadapter/dispatchermanager/dispatcher_manager.go
hongyunyan: It moves into runRemoveChangefeedCleanup, just like remove ddl_ts record.

lidezhu (approved):

PR #4840, opened 2026-04-16
eventservice: decouple dispatcher ready/handshake from scan triggering and activate the control plane immediately after register/reset

### LINKED ISSUE #4873: event service does not send ready/handshake immediately for newly registered or reset dispatchers when no new resolved-ts arrives, causing dispatcher initialization or recovery to stall
### What did you do?

Registered a new dispatcher, or reset an existing dispatcher after it had reached the ready stage, while keeping the upstream idle so that no new events advanced the resolved-ts after the register/reset.

The event service should send a `ReadyEvent` immediately after a dispatcher is registered successfully, and send a `HandshakeEvent` immediately after a reset switches the dispatcher to a new epoch, so that the event collector can complete the register/reset control-plane transition without waiting for new upstream traffic.

### What did you see instead?

Both `ReadyEvent` and `HandshakeEvent` were coupled to the `onNotify() -> scanReady()` path. When no new resolved-ts arrived:

- the collector never received `ReadyEvent` during initial registration, so it could not issue the follow-up reset request;
- after reset, the new epoch never received `HandshakeEvent`, so the collector could not initialize the new event stream.

As a result, the dispatcher could remain stuck in an uninitialized or post-reset unhandshaked state until another upstream resolved-ts happened to arrive.

### Versions of the cluster

master

zier-one:

Issue Number: close #4873

This PR fixes a control-plane timing issue in the event service. For normal dispatchers, `ReadyEvent` after registration and `HandshakeEvent` after reset were still dependent on a later resolved-ts notification.

This PR extracts dispatcher control-plane activation into `activateDispatcherControlPlane()`, which is responsible for:

- sending `ReadyEvent` for dispatchers in `epoch == 0`;
- sending `HandshakeEvent` for dispatchers that are ready but not handshaked yet.

The broker now calls this helper proactively at the key transition points instead of waiting for the next scan/notifier trigger:

- immediately after a normal dispatcher is registered successfully;
- immediately after a reset installs the new epoch;
- from the table-trigger tick path as the shared control-plane helper, while preserving existing behavior.

`scanReady()` still uses the same helper, so the existing scan path remains intact. The change only removes the hidden coupling that required another resolved-ts advance before the dispatcher could enter ready/handshake.

The PR also adds regression tests covering immediate ready after registration, immediate handshake after reset, and unchanged table-trigger registration behavior.
#### Tests 

- Unit test

### Release note 

```release-note


Fix a stall where a dispatcher could remain stuck after registration or reset when no new upstream resolved-ts arrived. The event service now emits the required ready/handshake control-plane events immediately instead of waiting for a later scan notification.

```

## Release Notes

* **Bug Fixes**
  * Fixed error handling during dispatcher initialization to ensure proper cleanup when registration fails.
  * Improved reliability of event sequencing during dispatcher lifecycle operations.

* **Tests**
  * Expanded test coverage for event emission timing and dispatcher state transitions.
  * Hardened tests for edge cases in schema synchronization failures.

### DIFF
--- a/pkg/eventservice/event_broker.go
+++ b/pkg/eventservice/event_broker.go
@@ -342,10 +342,9 @@ func (c *eventBroker) tickTableTriggerDispatchers(ctx context.Context) error {
 		case <-ticker.C:
 			c.tableTriggerDispatchers.Range(func(key, value interface{}) bool {
 				stat := value.(*atomic.Pointer[dispatcherStat]).Load()
-				if !c.checkAndSendReady(stat) {
+				if !c.activateDispatcherControlPlane(stat) {
 					return true
 				}
-				c.sendHandshakeIfNeed(stat)
 				startTs := stat.sentResolvedTs.Load()
 				remoteID := node.ID(stat.info.GetServerID())
 				keyspaceMeta := common.KeyspaceMeta{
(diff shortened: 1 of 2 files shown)


lidezhu:
1. The Ready event is used to indicate that the incremental scan has finished. We deliberately use it to avoid having too many incremental scan tasks running at the same time.
2. It is OK to send the Handshake event immediately after a reset.

PR #4950, opened 2026-04-29
eventservice: optimize scanwindow

### LINKED ISSUE #5041: scan window algorithm is unstable
The current scan window reacts too aggressively to downstream memory feedback. A single release pulse can reset the window to its default size, while later pressure reports quickly shrink it again. Since the scan window is a per-changefeed commit-ts span cap rather than a timer, this rapid expansion and contraction creates sawtooth behavior in scan batching. In practice, it causes throughput fluctuation and repeated memory pressure oscillation. In multi-dispatcher scenarios, lagging dispatchers can also pin the global base ts, amplifying instability and delaying progress for tables blocked by pending DDLs.

asddongmen:

Issue Number: close #5041 

The old scan window controller overreacted to memory feedback. A release pulse could reset the window, then later reports would shrink it again. Because it is a per-changefeed commit-ts span cap rather than a timer, this created sawtooth batching. Stale dispatchers could also pin the base ts and delay tables blocked by pending DDL.

This PR replaces the old policy with an adaptive controller while keeping scans event-driven. It combines a sliding usage window, EMAs, a pressure score, and cooldowns. High or critical pressure reduces the window in bounded steps. Low pressure recovers gradually and can leave the default floor faster. Release signals now relieve pressure instead of resetting the interval. The broker also skips stale dispatchers in `minSentTs` refresh and allows local advance for pending DDL when the global window is pinned.

This smooths scan progress, reduces reset storms, and preserves DDL forward progress.

**Before**
image

**After**
image

#### Tests 

- Unit test

No compatibility break is expected. This changes only scan window control behavior and should stabilize throughput and memory pressure.

### Release note 

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  * Added scan-window monitoring metrics for usage, EMAs, pressure, target bands, adjustment counts and memory-release events.

* **Improvements**
  * Replaced trend-based scan-interval logic with an adaptive, pressure-aware scan-window controller enabling tiered adjustments, braking, and conservative recovery.

* **Bug Fixes**
  * Scan interval no longer resets on memory-release; scan-window metrics are initialized and cleaned up consistently.

* **Tests**
  * Expanded and reworked tests covering adaptive behavior, braking/recovery, metrics, cooldowns and interval continuity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pingcap/ticdc/pull/4950)


3AceShowHand:
### Code review

Found 3 issues:

1. **Race condition in `observeScanWindowTargetBandMetrics`** — non-atomic Load-then-Store on band-state atomics causes double-counting of target-band cross events under concurrent calls. `handleCongestionControl` (event_broker.go:1320) iterates `changefeedMap.Range` and can be invoked concurrently from different event-collector nodes, so two goroutines can both observe the same `previousState` before either stores. Use `state.Swap(int32(currentState))` instead of `Load()` followed by `Store()`.

https://github.com/asddongmen/tigate/blob/260cb3237a6b3d504c5ca2129a76d1fd16f35a34/pkg/eventservice/scan_window.go#L417-L425

2. **Stale comments on threshold constants** — Three constant doc comments describe the old trend-based controller behavior and do not match the new adaptive controller:

   - `memoryUsageHighThreshold` (line 51): says reduction is to "1/2", but `shouldReduceForHighPressureLocked` uses `scaleDuration(current, 3, 4)` (3/4).
   - `memoryUsageLowThreshold` (line 62): says increase is uniformly "25%", but `scanWindowLowRecoveryScale` uses tiered scaling (5%/10%/25%) and additional EMA conditions from `isLowPressureLocked`.
   - `memoryUsageVeryLowThreshold` (line 74): says increase is uniformly "50%", but `scanWindowVeryLowRecoveryScale` uses tiered scaling (10%/20%/50%) and additional EMA conditions from `isVeryLowPressureLocked`.

https://github.com/asddongmen/tigate/blob/260cb3237a6b3d504c5ca2129a76d1fd16f35a34/pkg/eventservice/scan_window.go#L49-L78

3. **Brake evaluation before pressure relief on memory release** — In `OnCongestionReport`, the critical-brake and high-pressure checks run before `relievePressureLocked` is called. On a memory-release event, the release sample is added to the usage window and EMAs are updated first. If the slow EMA was at 0.90 and the release brings `usageRatio=0.20`, the slow EMA decays to 0.76 — still above `scanWindowHighPressureThreshold` (0.75) — causing a high-pressure reduction to fire during what should be a pressure-relief event. The old code handled this by resetting to default and returning early. Consider applying relief before evaluating non-critical downward adjustments, or skipping downward adjustments when `memoryReleaseCount > 0`.

https://github.com/asddongmen/tigate/blob/260cb3237a6b3d504c5ca2129a76d1fd16f35a34/pkg/eventservice/scan_window.go#L458-L479

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

3AceShowHand (approved):

lidezhu (approved):

PR #5006, opened 2026-05-07
mysql,sqlmodel: support table route in mysql sink

### LINKED ISSUE #4818: integrate the table route functionality into the mysql sink
mysql sink can works well after enable the table route

3AceShowHand:

Issue Number: close #4818

This PR makes the MySQL sink honor table route rules when generating and
executing downstream SQL. Before this change, parts of the MySQL sink still used
source schema/table names directly, so a routed changefeed could generate DML or
execute DDL against the source database/table instead of the routed downstream
target.

- Use routed schema names when executing DDL in the MySQL sink.
  - `needSwitchDB` checks `DDLEvent.GetTargetSchemaName()`.
  - `execDDL` runs `USE <target schema>` when the DDL needs a default database.

- Use routed table names when generating MySQL DML SQL.
  - MySQL insert/delete/update and active-active upsert SQL use the target table
    name stored in `TableInfo`.
  - `sqlmodel` single-row and multi-row insert/delete/update builders use
    `QuoteTargetString()`, so the same row model can generate SQL for routed
    targets.

- Keep routed metadata usable across DDL and DML paths.
  - DDL rewrite fills the DDL event default schema before applying routing, so
    unqualified table names in DDL can be routed correctly.
  - Routed `TableInfo` clones initialize their private SQL fields before being
    used by downstream sinks.
  - The event collector avoids replacing a dispatcher's cached table info with
    unrelated DDL table info, while still advancing the delivered table info
    version for DDLs sent to that dispatcher.

- Add a MySQL-only table route integration test.
  - The test routes `source_db.*` to `target_db.*_routed` and
    `source_extra_db.*` to `target_extra_db.*_routed`.
  - It covers DML, create/alter/drop/rename/truncate table, create/drop view,
    partition DDL, cross-database DDL, and drop database.
  - It validates routed data with sync-diff route rules and checks that dropped
    routed databases disappear downstream.

#### Tests 

- Unit test
- Integration test

No compatibility break is expected. Without table route rules, target schema and
table names fall back to the original source names.

No meaningful performance regression is expected. The MySQL sink uses target
names already carried by routed events/table info instead of applying routing in
the DML hot path.

No. This PR wires existing table route behavior into the MySQL sink and adds
integration coverage; it does not introduce new user-facing configuration or
monitoring items.

### Release note 

```release-note
Support table route for the MySQL sink.
```


downstreamadapter/eventcollector/dispatcher_stat.go
```diff
@@ -360,10 +360,6 @@ func (d *dispatcherStat) handleBatchDataEvents(events []dispatcher.DispatcherEve
 			batchDML := event.Event.(*commonEvent.BatchDMLEvent)
 			batchDML.AssembleRows(tableInfo)
 			for _, dml := range batchDML.DMLEvents {
-				// DMLs in the same batch share the same updateTs in their table info,
-				// but they may reference different table info objects,
-				// so each needs to be initialized separately.
-				dml.TableInfo.InitPrivateFields()
```
wk989898: Why remove this?

3AceShowHand: It's called when the table info is initialized, should not be called here.

lidezhu: I think it cannot be removed becuase it should be used in remote read mode.

wk989898 (approved):

hongyunyan (approved):

PR #5053, opened 2026-05-15
redo, event: preserve table route target names in redo log

### LINKED ISSUE #4819: integrate the table route into the redo log
The redo log functionality can works well after enable the table route

3AceShowHand:

Issue Number: close #4819

When table route is configured and redo log is enabled, DDL events without a `TableInfo` (e.g., `CREATE DATABASE`) lose their routing target names in the redo log. During redo apply, the replayed DDL cannot be routed to the correct downstream target schema. DML and DDL events with `TableInfo` also relied on `TableInfo.GetTargetSchemaName()` but were inconsistent with how the event-level target names were set by the dispatcher.

**Core fix** (`pkg/common/event/redo.go`):

- `DDLEvent.ToRedoLog()`: Now uses `d.GetTargetSchemaName()` / `d.GetTargetTableName()` to populate the redo log's `TableName`. These methods return the event-level target names set by the table route dispatcher (via `NewRoutedDDLEvent`), which works correctly even when `TableInfo` is nil. Previously the code only read from `TableInfo`, missing schema-level DDL events entirely.

- `RedoRowEvent.ToRedoLog()`: Uses `r.TableInfo.GetTargetSchemaName()` / `GetTargetTableName()`, which correctly return routed names when `TableInfo` has `TargetSchema`/`TargetTable` set.

- Both methods refactored for readability: intermediate variables replace deeply nested struct initialization, and `nil` TableInfo cases return early.

**Other changes**:

- `pkg/common/event/ddl_event.go`: Removed unnecessary nil-check in `GetDDLQuery()`.
- `downstreamadapter/sink/redo/sink.go`: Minor cleanup.
- `pkg/common/event/redo_test.go`: Added unit tests covering DDL/DML without routing, DML with nil TableInfo, and enhanced existing test assertions.

**Integration test** (`tests/integration_tests/redo_apply_table_route/`):

Validates end-to-end correctness: creates a changefeed with table route config, runs a complex DDL/DML workload, stops the sink writer with a failpoint to accumulate redo logs, applies redo, and verifies all replayed data lands in the correct routed downstream tables.

#### Tests

- Unit test
- Integration test

No. The refactoring preserves identical logic. Redo log format is unchanged — only the values written to existing fields differ when routing is active.

No.

### Release note

```release-note
Integrate  the redo into the table route
```

### DIFF
--- a/downstreamadapter/sink/redo/sink.go
+++ b/downstreamadapter/sink/redo/sink.go
@@ -159,7 +159,7 @@ func (s *Sink) WriteBlockEvent(event commonEvent.BlockEvent) error {
 		}
 		log.Info("redo sink send DDL event",
 			zap.String("keyspace", s.changefeedID.Keyspace()), zap.String("changefeed", s.changefeedID.Name()),
-			zap.Any("event", e.GetDDLQuery()), zap.Any("startTs", event.GetStartTs()), zap.Any("commitTs", event.GetCommitTs()),
+			zap.Any("query", e.GetDDLQuery()), zap.Any("startTs", event.GetStartTs()), zap.Any("commitTs", event.GetCommitTs()),
 			zap.String("schema", e.GetSchemaName()), zap.String("table", e.GetTableName()), zap.Int64("tableID", e.GetTableID()))
 	}
 	return nil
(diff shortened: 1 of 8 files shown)


downstreamadapter/sink/redo/sink.go
```diff
@@ -159,7 +159,7 @@ func (s *Sink) WriteBlockEvent(event commonEvent.BlockEvent) error {
 		}
 		log.Info("redo sink send DDL event",
 			zap.String("keyspace", s.changefeedID.Keyspace()), zap.String("changefeed", s.changefeedID.Name()),
-			zap.Any("event", e.GetDDLQuery()), zap.Any("startTs", event.GetStartTs()), zap.Any("commitTs", event.GetCommitTs()),
+			zap.Any("query", e.GetDDLQuery()), zap.Any("startTs", event.GetStartTs()), zap.Any("commitTs", event.GetCommitTs()),
```
wk989898:
```suggestion
			zap.String("query", e.GetDDLQuery()), zap.Any("startTs", event.GetStartTs()), zap.Any("commitTs", event.GetCommitTs()),
```

wk989898 (approved):

lidezhu (approved):

tenfyzhong (approved):

PR #5054, opened 2026-05-15
pulsar: support debezium protocol

### LINKED ISSUE #5056: pulsar: support debezium protocol

The Pulsar sink only accepts `canal-json` as the message protocol. Users who consume TiCDC change events via Pulsar with Debezium-compatible downstream systems (e.g. Flink CDC, Debezium connectors) have no way to produce standard Debezium-format messages, even though the Debezium codec is already implemented and used by the Kafka sink.

Allow `protocol=debezium` to be set in the Pulsar sink URI:

  pulsar://host:6650/topic?protocol=debezium

This requires extending `IsPulsarSupportedProtocols()` in `pkg/config/sink_protocol.go` to include `ProtocolDebezium`. No new codec logic is needed since the Debezium encoder/decoder in `pkg/sink/codec/debezium/` is already shared with the Kafka sink via the common codec builder.

Users currently must use `canal-json` and implement a custom transform layer in their consumer to convert to Debezium format. This adds operational overhead and is error-prone.

The Pulsar sink URI parameter documentation should be updated to list `debezium` as a valid value for the `protocol` parameter alongside `canal-json`. No migration is required for existing changefeeds since the default protocol is unchanged.

HGHNice:

Issue Number: close #5056 

The Pulsar sink currently only supports `canal-json`. Users who consume
TiCDC events via Pulsar with Debezium-compatible consumers (e.g. Flink CDC)
have no way to use the standard Debezium message format.

- Extend `IsPulsarSupportedProtocols()` in `pkg/config/sink_protocol.go`
  to include `ProtocolDebezium`.
- Update the error message in `downstreamadapter/sink/pulsar/helper.go`
  to reflect the expanded protocol list.
- The Debezium codec is already fully implemented and shared with the Kafka
  sink via the common codec builder, so no additional encoding logic is needed.
- Add unit test `TestIsPulsarSupportedProtocols`.

#### Tests

- [x] Unit test

No. Existing `canal-json` behavior is unchanged.

The Pulsar sink docs should note that `debezium` is now a valid `protocol` value.

### Release note

```release-note
The Pulsar sink now supports the `debezium` protocol. Set `protocol=debezium`
in the Pulsar sink URI to produce Debezium-format change events.


## Summary by CodeRabbit

## Release Notes

* **New Features**
  * Pulsar sink now supports debezium protocol in addition to canal-json protocol.

* **Tests**
  * Added validation tests for Pulsar sink protocol compatibility.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/pingcap/ticdc/pull/5054)

<!-- review_stack_entry_end -->
```

## Summary by CodeRabbit

* **New Features**
  * Pulsar sink now supports Debezium protocol format in addition to Canal-JSON.

* **Tests**
  * Added unit tests for protocol parsing and Pulsar-supported protocol validation.
  * Updated Debezium integration test flow to run against both Kafka and Pulsar sink types.

### DIFF
--- a/downstreamadapter/sink/pulsar/helper.go
+++ b/downstreamadapter/sink/pulsar/helper.go
@@ -111,7 +111,7 @@ func newPulsarSinkComponentWithFactory(ctx context.Context,
 		return pulsarComponent, protocol, errors.Trace(err)
 	}
 
-	pulsarComponent.eventRouter, err = eventrouter.NewEventRouter(sinkConfig, topic, true, false)
+	pulsarComponent.eventRouter, err = eventrouter.NewEventRouter(sinkConfig, topic, true, protocol == config.ProtocolAvro)
 	if err != nil {
 		return pulsarComponent, protocol, errors.Trace(err)
 	}
--- a/downstreamadapter/sink/pulsar/helper.go
+++ b/downstreamadapter/sink/pulsar/helper.go
@@ -88,7 +88,7 @@ func newPulsarSinkComponentWithFactory(ctx context.Context,
 	if !config.IsPulsarSupportedProtocols(protocol) {
 		return pulsarComponent, protocol, errors.ErrSinkURIInvalid.
 			GenWithStackByArgs("unsupported protocol, " +
-				"pulsar sink currently only support these protocols: [canal-json]")
+				"pulsar sink currently only support these protocols: [canal-json, debezium]")
 	}
 
 	pulsarComponent.config, err = pulsar.NewPulsarConfig(sinkURI, sinkConfig.PulsarConfig)
@@ -111,7 +111,6 @@ func newPulsarSinkComponentWithFactory(ctx context.Context,
 		return pulsarComponent, protocol, errors.Trace(err)
 	}
 
-	// pulsar only support canal-json, so we don't need to check the protocol
 	pulsarComponent.eventRouter, err = eventrouter.NewEventRouter(sinkConfig, topic, true, false)
 	if err != nil {
 		return pulsarComponent, protocol, errors.Trace(err)
--- a/pkg/config/sink_protocol.go
+++ b/pkg/config/sink_protocol.go
@@ -180,5 +180,5 @@ func GetScheme(url *url.URL) string {
 
 // IsPulsarSupportedProtocols returns whether the protocol is supported by pulsar.
 func IsPulsarSupportedProtocols(p Protocol) bool {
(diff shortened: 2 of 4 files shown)


wk989898:
Please add the Debezium integration tests for Pulsar.

HGHNice:
> Please add the Debezium integration tests for Pulsar.


Pulsar Debezium integration tests have been added. Thank！

wk989898 (approved):

3AceShowHand (approved):

flowbehappy (approved):

HGHNice:
Hi @wk989898, I dug into the failing pull-cdc-pulsar-integration-heavy job. The new debezium_basic pulsar case fails because the data never reaches the downstream — the pulsar consumer panics while decoding. After tracing it, I found this is a real gap in debezium-over-pulsar, not just a test issue:

Root cause

The common.Message carries both Key []byte (the encoded message key — for debezium this is the debezium key JSON) and PartitionKey *string (used only for pulsar routing).
The kafka sink sends Key: message.Key, so the kafka consumer decodes the debezium key correctly.
The pulsar sink instead sends Key: message.GetPartitionKey() (dml_producer.go / ddl_producer.go), so the encoded message.Key is dropped. The pulsar consumer then feeds the routing key into debezium.decoder.AddKeyValue, which calls decodeRawBytes on it, fails (it isn't valid debezium JSON), and log.Panics — so nothing is written downstream and finish_mark1 never appears.
canal-json is unaffected because its decoder ignores the key (AddKeyValue(_, value)), which is why the existing pulsar canal-json case passes.
So enabling protocol=debezium lets the pulsar sink produce debezium messages, but the debezium key is currently not delivered over pulsar, which breaks any debezium consumer that relies on the key (not just our test tool).

Possible directions

Make the pulsar sink carry the encoded message.Key (e.g. put it on the pulsar message Key and use OrderingKey for routing), and read it back in the consumer. This is the "correct" fix but changes the produced pulsar message format.
Make the debezium decoder tolerate a missing/non-debezium key and derive the primary key from the value schema instead. Smaller, but touches the shared decoder used by kafka too.
Keep this PR limited to enabling the protocol and address the end-to-end key delivery in a follow-up.
Could you let me know which direction you'd prefer? Happy to implement it.

PR #5194, opened 2026-06-03
filter: support ignore update only columns for kafka sink

### LINKED ISSUE #5269: filter: support ignore update only columns for kafka sink

In Materialized View (MV) + CDC synchronization scenarios, some upstream tables are periodically upserted with bookkeeping columns such as version or updated_at. In many cases, business columns remain unchanged while only those bookkeeping columns are updated. Forwarding such UPDATE events still creates downstream traffic, but does not provide additional business value.
For example:
- A row has columns: id, name, gender, version, updated_at
- A periodic upsert updates only version and updated_at
- name and gender remain unchanged
- This still produces CDC UPDATE events and increases downstream processing cost

Configuration is added under the existing filter.event-filters section:
```
[[filter.event-filters]]
matcher = ["test.user"]
ignore-update-only-columns = ["version", "updated_at"]
```

#### Behavior
Matching Logic:
- For each UPDATE event, determine the set of columns whose values changed.
- If all changed columns are included in ignore-update-only-columns, ignore the event.
- If any changed column is not included in ignore-update-only-columns, keep the event and sink it downstream.2. Column 

Resolution: 
- Column names must match the schema column names exactly. The casing behavior follows the case-sensitive configuration.
- If a configured column does not exist in the table schema, log a warning and ignore that entry.

Event Types:
- Only applies to UPDATE events.
- INSERT and DELETE events are not affected.

Downstream Scope:
- In the current iteration, this configuration is honored only when the downstream sink is Kafka.

_No response_

_No response_

lidezhu:

Issue Number: close #5269 

This pull request introduces a value-aware filtering capability for the Kafka sink in TiCDC. By allowing users to specify columns that should be ignored during update events, the system can now suppress unnecessary downstream updates when only non-critical columns (like metadata or version fields) are modified. This change involves updates to the configuration model, the event filtering engine, and the dispatcher logic to propagate these settings effectively.

### Highlights

* **Feature Implementation**: Added support for filtering out update events in the Kafka sink when only ignored columns are modified.
* **Configuration**: Introduced `ignore-update-only-columns` configuration to specify columns that should not trigger an update event if they are the only ones changed.

#### Tests 

- Unit test
- Integration test
- No code

### Release note 

```release-note

```

## Summary by CodeRabbit

* **New Features**
  * Added an API and configuration option to ignore UPDATE events when only specified “update-only” columns change.
  * Extended dispatcher requests and behavior flags so the feature is enabled for Kafka sinks.
* **Tests**
  * Added/updated unit tests covering the dispatcher capability toggle and update-only-column filtering logic.
  * Added an integration test (with dedicated changefeed config and validation) to verify end-to-end output for Kafka-based raw change events.


lidezhu:
@coderabbitai review

asddongmen (approved):

pkg/filter/update_only_columns_filter.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	for _, col := range tableInfo.GetColumns() {
+		if strings.EqualFold(col.Name.O, name) {
+			return col, true
+		}
+	}
+	return nil, false
+}
+
+func (r *updateOnlyColumnsRule) warnInvalidColumn(tableInfo *common.TableInfo, columnName string) {
```
3AceShowHand: no need to have a method for logging, inline it should the concret line number.

3AceShowHand: This looks a verbose log.

pkg/config/filter.go
```diff
@@ -39,8 +39,9 @@ type EventFilterRule struct {
...
 	// sql expression
-	IgnoreInsertValueExpr    *string `toml:"ignore-insert-value-expr" json:"ignore-insert-value-expr,omitempty"`
-	IgnoreUpdateNewValueExpr *string `toml:"ignore-update-new-value-expr" json:"ignore-update-new-value-expr,omitempty"`
-	IgnoreUpdateOldValueExpr *string `toml:"ignore-update-old-value-expr" json:"ignore-update-old-value-expr,omitempty"`
-	IgnoreDeleteValueExpr    *string `toml:"ignore-delete-value-expr" json:"ignore-delete-value-expr,omitempty"`
+	IgnoreInsertValueExpr    *string  `toml:"ignore-insert-value-expr" json:"ignore-insert-value-expr,omitempty"`
+	IgnoreUpdateNewValueExpr *string  `toml:"ignore-update-new-value-expr" json:"ignore-update-new-value-expr,omitempty"`
+	IgnoreUpdateOldValueExpr *string  `toml:"ignore-update-old-value-expr" json:"ignore-update-old-value-expr,omitempty"`
+	IgnoreDeleteValueExpr    *string  `toml:"ignore-delete-value-expr" json:"ignore-delete-value-expr,omitempty"`
+	IgnoreUpdateOnlyColumns  []string `toml:"ignore-update-only-columns" json:"ignore-update-only-columns,omitempty"`
```
3AceShowHand, line 46: This looks can be configured when using non kafka sink.

It should be rejected or print warn log

downstreamadapter/dispatcher/basic_dispatcher_info.go
```diff
@@ -195,6 +195,10 @@ func (d *BasicDispatcher) IsOutputRawChangeEvent() bool {
 	return d.sharedInfo.outputRawChangeEvent
 }
 
+func (d *BasicDispatcher) EnableIgnoreUpdateOnlyColumns() bool {
+	return d.sink.SinkType() == common.KafkaSinkType
```
3AceShowHand, line 199: It looks should also consider configuration.

pkg/filter/update_only_columns_filter.go
lidezhu: Fixed. And this log will only be printed when there is a new version of the table info, so it is not verbose.

pkg/config/filter.go
lidezhu: Add a warn log.

downstreamadapter/dispatcher/basic_dispatcher_info.go
lidezhu: This just means whether we need respect the config `IgnoreUpdateOnlyColumns` for this dispacter.

3AceShowHand (approved):

nongfushanquan (approved):

PR #5271, opened 2026-06-09
config: make minSyncPointInterval configurable via debug config

### LINKED ISSUE #5270: feature: make minSyncPointInterval configurable via debug config
## Is your feature request related to a problem?

The `minSyncPointInterval` is hardcoded to 30 seconds in `pkg/config/replica_config.go` (line 36):

```go
minSyncPointInterval = time.Second * 30
```

This value is enforced during `ValidateAndAdjust()` — any `SyncPointInterval` smaller than 30s is rejected:

```go
if c.SyncPointInterval != nil &&
    *c.SyncPointInterval < minSyncPointInterval {
    return cerror.ErrInvalidReplicaConfig.FastGenByArgs(...)
}
```

For integration tests that exercise the sync point feature, waiting 30 seconds per sync point cycle makes tests unnecessarily slow and hurts developer productivity.

## Describe the feature you'd like

Make `minSyncPointInterval` configurable through the existing `DebugConfig` in `pkg/config/debug.go`, similar to how other internal knobs (e.g., puller resolved-ts-stuck-interval, scheduler settings) are already exposed as debug config.

Proposed approach:
- Add a field like `MinSyncPointInterval` to a new or existing section in `DebugConfig` (e.g., under a `Replica` or `Changefeed` debug config section).
- In `ValidateAndAdjust()`, read the debug config value instead of the hardcoded constant. If the debug config is not set, fall back to the current default of 30s.
- The debug config is unexposed (not part of the public API), so this change has zero impact on production users.

This allows integration tests to set a smaller minimum (e.g., 1s–5s) via the debug config, significantly reducing test execution time.

## Describe alternatives you've considered

- **Override at test level only**: Could pass a flag or environment variable, but leveraging the existing debug config mechanism is cleaner and consistent with how other internal parameters are tuned.
- **Lower the global constant**: Not acceptable — the 30s minimum is a safety guard for production workloads to prevent excessive sync point generation overhead.

## Teachability, Documentation, Adoption, Migration Strategy

- No user-facing changes — this is purely a debug/internal configuration.
- Integration test authors can set the debug config in test fixtures to reduce `minSyncPointInterval` and speed up sync-point-related test cases.
- The pattern already exists in the codebase: `DebugConfig` hosts puller, scheduler, event-store, and schema-store knobs. Adding a replica/changefeed section follows the same convention.

pingyu:

Issue Number: close #5270

Make `minSyncPointInterval` configurable through the existing `DebugConfig` in `pkg/config/debug.go`, instead of being hardcoded to 30 seconds.

Changes:
- **`pkg/config/debug.go`**: Add `DebugReplicaConfig` with `MinSyncPointInterval` field under `debug.replica-config`. The configurable range is [5s, 30s]. Add `getMinSyncPointInterval()` to read the value from global server config with fallback to 30s default.
- **`pkg/config/replica_config.go`**: Replace hardcoded `minSyncPointInterval` constant with `minSyncPointIntervalDefault` (30s) and `minConfigurableSyncPointInterval` (5s). Split `ValidateAndAdjust` into `ValidateAndAdjust` (server-side, uses debug config minimum) and `ValidateAndAdjustForCLI` (client-side, uses absolute minimum 5s).
- **`pkg/config/server.go`**: Add `ReplicaConfig: NewDefaultDebugReplicaConfig()` to default server config.
- **`cmd/cdc/cli/cli_changefeed_create.go`**: Add `--sync-point` and `--sync-interval` CLI flags. Use `ValidateAndAdjustForCLI` for CLI-side validation.
- **`tests/integration_tests/syncpoint/`**: Replace changefeed.toml (which set sync-point-interval=30s) with server.toml debug config (min-sync-point-interval=5s), and use `--sync-point --sync-interval 10s` flags instead.

This allows integration tests to set a smaller minimum (e.g., 5s) via the debug config, significantly reducing test execution time while keeping the 30s safety guard for production.

#### Tests

- Unit test
- Integration test

No. The default minimum remains 30s. Only when explicitly configured via debug config (unexposed/internal) does the minimum change.

No. This is purely a debug/internal configuration with no user-facing changes.

### Release note

```release-note
config: minSyncPointInterval is configurable via debug config
```

## Summary by CodeRabbit

* **New Features**
  * CLI flags to enable sync points and set sync-point interval during changefeed creation.
  * New debug replica configuration to allow a lower, configurable minimum sync-point interval.

* **Validation**
  * Enforced distinct minimums for sync-point intervals: stricter global default and a more permissive CLI/debug minimum with clear validation errors.

* **Tests**
  * Added unit and integration tests covering flags, interval limits, config decoding, and end-to-end sync-point behavior.


tenfyzhong (approved):

lidezhu:
I think adding this configuration just for testing is not a good idea—it introduces unnecessary complexity. You might as well just hack the value directly.

PR #5420, opened 2026-06-16
kafka: decouple batch size from Kafka message size limit

### LINKED ISSUE #1405: Don't raise errors for the events larger than max-message-bytes
Currently, TiCDC use `max-message-bytes` for two usages:
1. To be the max batch size during batching multiple events into one message before sending to Kafka
2. Stop message larger than `max-message-bytes` being sent to Kafka. CDC raises an error if a message is larger, e.g. `Single message is too large for avro`.

While, in fact, those two usages should be set separately, instead of relying on one single setting. Because a user could send a single event with larger size.

And since there is already a `message.max.bytes` in the Kafka topic setting or Kafka Broker setting, we don't need to limit the message size on TiCDC any more. So I suggest we simply remove the second usage of `max-message-bytes` from TiCDC.

3AceShowHand:

Issue Number: close #1405

Kafka sink currently uses the changefeed `max-message-bytes` setting for two different purposes:

- deciding when TiCDC should split a batch of row events;
- limiting the final Kafka message and deciding when to invoke `large-message-handle`.

This means TiCDC can reject a message even when Kafka can accept it. For example, if the changefeed has `max-message-bytes=10 MiB`, the Kafka topic allows `20 MiB`, and one encoded row event is `12 MiB`, TiCDC still returns `ErrMessageTooLarge` or invokes `large-message-handle`. Operators must then pause the changefeed and update its configuration in addition to updating Kafka.

Whether a message is too large should be determined by Kafka's message-size limit. The changefeed setting should only control TiCDC's batching behavior.

This PR separates the Kafka message limit from the TiCDC batch threshold. Define:

- `T`: the changefeed `max-message-bytes` setting;
- `K`: the Kafka topic `max.message.bytes` for an existing topic, or the broker `message.max.bytes` for a topic that does not exist yet;
- producer message limit: `K`;
- batch threshold: `min(T, K)`.

The implementation:

- preserves `T` as `MaxBatchedBytes` before reading Kafka metadata;
- updates `MaxMessageBytes` to `K` and uses it for both the producer and the codec's final message-size check;
- passes both limits to codec configuration;
- uses `MaxBatchedBytes` only when deciding whether to add another row to the current batch;
- invokes `large-message-handle` only when the original encoded message exceeds `MaxMessageBytes` (`K`), and performs a final size check after handling;
- falls back to the configured `T` as the producer limit if the Kafka topic or broker size configuration cannot be read.

As a result:

- a single message larger than `T` but not larger than `K` is sent directly;
- multiple rows are still split according to `min(T, K)`;
- a message larger than `K` invokes the configured `large-message-handle`, or returns `ErrMessageTooLarge` when it still cannot fit;
- after the Kafka size limit is increased, a restarted Kafka sink reads the new limit and can recover without changing the changefeed `max-message-bytes` setting.

This PR does not introduce a separate model for Sarama's request-size limit or cap `K` using `sarama.MaxRequestSize`. Sarama request sizing remains outside the scope of this change.

Non-Kafka sinks pass the same value for the final message limit and batch threshold, preserving their existing behavior.

#### Tests

- Unit test
  - Kafka topic and broker limits are propagated to the producer and codec configuration.
  - The batch threshold remains `min(T, K)`.
  - Open Protocol splits batches at the batch threshold while allowing one row larger than that threshold when it is within the final message limit.
  - Open Protocol, Canal JSON, and Simple Protocol enforce the final message limit and large-message handling boundary.
- Integration test
  - `kafka_big_messages` covers Canal JSON, Open Protocol, Simple JSON, Simple Avro, and Avro.
  - Each case first sends a message larger than Kafka's current topic limit and verifies `ErrMessageTooLarge`.
  - The test then only increases the topic `max.message.bytes`, without updating, pausing, or manually resuming the changefeed, and verifies that synchronization automatically recovers and passes sync-diff.
  - Existing claim-check and handle-key-only cases explicitly configure the Kafka topic limit so they continue to exercise the large-message handler.

No performance regression or compatibility break is expected.

The batching behavior of `max-message-bytes` is preserved. This PR fixes the previous conflation of the batch threshold and Kafka's actual message-size limit: when `T < message size <= K`, Kafka can accept the message, so TiCDC sends it instead of incorrectly returning `ErrMessageTooLarge` or invoking `large-message-handle`. If `K < T`, the batch threshold is reduced to `K` so TiCDC does not intentionally build an unsendable batch.

Yes. User documentation should clarify that `max-message-bytes` controls Kafka batching, while Kafka topic or broker configuration controls the final message limit. No monitoring update is required.

### Release note

```release-note
Decouple the Kafka sink batch threshold from Kafka's message-size limit, so increasing the Kafka topic or broker limit can unblock a large row event without changing the TiCDC changefeed configuration.
```


pkg/sink/kafka/sarama_config.go
```diff
@@ -62,7 +62,7 @@ func newSaramaConfig(ctx context.Context, o *options) (*sarama.Config, error) {
 	config.Producer.Flush.Bytes = 0
 	config.Producer.Flush.Messages = 0
 	config.Producer.Flush.Frequency = time.Duration(0)
-	config.Producer.Flush.MaxMessages = o.MaxMessages
+	config.Producer.Flush.MaxMessages = 0
```
wk989898, line 65: Why update this?

pkg/sink/codec/common/config.go
```diff
@@ -41,8 +41,8 @@ type Config struct {
 
 	Protocol config.Protocol
 
-	// control batch behavior, only for `open-protocol` and `craft` at the moment.
 	MaxMessageBytes int
+	MaxBatchedBytes int
```
wk989898, line 47: Please add a comment to explain why this parameter was introduced?

pkg/sink/kafka/sarama_config.go
3AceShowHand: this is a dead not used value.

wk989898 (approved):

lidezhu (approved):

Issue #5083, opened 2026-05-18
Duplicate dispatcher can be created during maintainer failover before orphan dispatcher drains


asddongmen:
During maintainer failover, TiCDC may create duplicate dispatchers for the same table span and `startTs`. One dispatcher is owned by the new maintainer, while another appears to be a delayed/orphaned dispatcher request from the previous maintainer. The orphan dispatcher is removed after the new maintainer observes it, but it may already have entered `Working` state and pushed DMLs into the sink, causing downstream write conflicts.

In the incident around `2026-05-11 23:08:45`, new maintainer `fba6e585-1710-4b74-9102-fcae45514fff` bootstrapped with `nodeCount=1`, `spanCount=17`, and checkpoint/startTs `466233824429998495`.

For tableID `121` (`workload.sbtest20`), the current maintainer created dispatcher `36949683455125699276247693423456213681`. Another dispatcher, `106138415981049979221359258994415068339`, was later created for the same full span and same startTs, but no corresponding maintainer span/operator was found. Maintainer then logged `no span found, remove it`, while the dispatcher still had `tableProgressLen=378`.

For tableID `142` (`workload.sbtest27`), the same pattern happened between dispatcher `354956623857882180211263781318113896091` and orphan dispatcher `1172258264747904389658988347929528985`; the orphan had `tableProgressLen=380` when removal started.

Shortly after both duplicated dispatcher pairs handshook, TiCDC reported downstream `Error 9007 Write conflict` and retried DMLs.

Expected behavior: for one changefeed/mode/table span, only one active dispatcher should be able to write to the downstream sink during failover.

Suspected cause: bootstrap only reconstructs state from alive dispatcher managers. A delayed create request from the previous maintainer can still be processed by dispatcher manager after the new maintainer has already recreated the same table span. There is no global fence preventing the orphan dispatcher from writing before it is detected and drained.

asddongmen:
related logs:

[20260518.zip](https://github.com/user-attachments/files/27951487/20260518.zip)

PR #5434, opened 2026-06-18
coordinator: persist maintainer epochs before ownership changes

hongyunyan:

Issue Number: ref #5083

This is PR 1 of 3 split from #5182.

Background:

During maintainer failover, the coordinator can schedule a new maintainer while
delayed requests from the previous maintainer are still in flight. Receiver-side
fencing only works if every new ownership attempt has a persisted, monotonic
maintainer epoch before any add or move request is sent.

Motivation:

The old code generated owner epochs from in-memory state and resumed changefeeds
with a normal metadata write. That made it hard to prove that a new maintainer
owner is always ordered after the previously persisted owner, especially across
coordinator failover, resume, retry, and move paths.

This PR adds the coordinator-side epoch persistence foundation:

- Adds optional maintainer epoch fields to maintainer lifecycle, scheduling,
  heartbeat, and close messages.
- Adds common maintainer epoch matching helpers and PD TSO based epoch generation.
- Adds `BumpChangefeedEpoch` to atomically persist a strictly newer changefeed
  epoch, optionally with status updates under the same etcd transaction.
- Uses the persisted epoch bump before add/move operators, resume, and warning
  retry scheduling can create a new maintainer owner.
- Keeps add-maintainer config bytes synchronized with the latest persisted
  `ChangeFeedInfo`.
- Makes coordinator bootstrap select the current-epoch maintainer and stop stale
  reported owners with their original epoch.
- Carries the origin epoch through stop and move operators so old owners are
  fenced by the epoch they actually own.

Stack:

- [#5434](https://github.com/pingcap/ticdc/pull/5434): coordinator epoch persistence and owner scheduling foundation.
- [#5435](https://github.com/pingcap/ticdc/pull/5435): maintainer and dispatcher-manager receiver-side epoch fence.
- [#5436](https://github.com/pingcap/ticdc/pull/5436): table trigger takeover hardening and integration coverage.

#### Tests 

- Unit test

No expected performance regression. Epoch bumps happen on ownership-changing
control paths such as add, move, resume, and retry scheduling, not in the event
write path.

The protobuf fields are optional. Epoch 0 remains the compatibility value used
by old components until later PRs enforce strict receiver-side fences.

No.

### Release note 

```release-note
None
```

### Validation

- `make generate-protobuf`
- `make fmt`
- `go test ./coordinator/changefeed ./coordinator ./coordinator/operator ./pkg/pdutil`

## Summary by CodeRabbit

* **New Features**
  * Added maintainer epoch propagation through heartbeat and maintainer lifecycle requests, including resume and stop flows.
* **Bug Fixes**
  * Prevented stale maintainer reports from being applied by dropping mismatched heartbeat status.
  * Updated bootstrap/stop/remove behavior to fence stale owners while preserving the reported maintainer epoch.
* **Improvements**
  * Tightened changefeed initialization/resume validation and improved warning-state handling by bumping ownership epoch before stopping.
* **Tests**
  * Expanded coverage for epoch bumping, resume behavior, and maintainer-epoch propagation.


pkg/pdutil/utils.go
```diff
@@ -66,3 +66,14 @@ func GenerateChangefeedEpoch(ctx context.Context, pdClient pd.Client) uint64 {
 	}
 	return oracle.ComposeTS(phyTs, logical)
 }
+
+// AdvanceChangefeedEpoch returns max(candidate, current+1).
+func AdvanceChangefeedEpoch(candidate, current uint64) (uint64, error) {
```
3AceShowHand: Changefeed epoch should not be a part of the pd util?

pkg/pdutil/utils.go
```diff
@@ -66,3 +66,14 @@ func GenerateChangefeedEpoch(ctx context.Context, pdClient pd.Client) uint64 {
...
+	if candidate > current {
+		return candidate, nil
+	}
+	if current == ^uint64(0) {
```
3AceShowHand: Will this really happen ?

coordinator/changefeed/changefeed.go
```diff
@@ -243,10 +234,11 @@ func (c *Changefeed) GetStatusForResume() *heartbeatpb.MaintainerStatus {
...
 	clone := &heartbeatpb.MaintainerStatus{
-		CheckpointTs: status.CheckpointTs,
-		FeedState:    status.FeedState,
-		State:        status.State,
-		// we don't clone the errors from status, because the old error is meaningless for the resume action, but only blocks.
+		CheckpointTs:    status.CheckpointTs,
+		FeedState:       status.FeedState,
+		State:           status.State,
+		MaintainerEpoch: status.MaintainerEpoch,
+		// Old errors are meaningless for resume and can only block the resumed task.
```
3AceShowHand: This comment lack of context and misleading.

coordinator/changefeed/changefeed.go
```diff
@@ -272,21 +264,26 @@ func (c *Changefeed) GetLastSavedCheckPointTs() uint64 {
 }
 
 func (c *Changefeed) NewAddMaintainerMessage(server node.ID) *messaging.TargetMessage {
+	info := c.GetInfo()
+	if info == nil {
```
3AceShowHand: info is nil should be checked when new the changefeed ?

coordinator/changefeed/changefeed.go
```diff
@@ -272,21 +264,26 @@ func (c *Changefeed) GetLastSavedCheckPointTs() uint64 {
...
+		log.Panic("changefeed info is nil", zap.String("changefeedID", c.ID.String()))
+	}
+	configData, err := info.MarshalWithTruncation(false)
+	if err != nil {
```
3AceShowHand, line 274: ditto.

coordinator/changefeed/changefeed_db_backend.go
```diff
@@ -30,14 +43,15 @@ type Backend interface {
...
+	// ChangeFeedInfo. It only reads and updates stored status when UpdateStatus is set.
+	BumpChangefeedEpoch(ctx context.Context, id common.ChangeFeedID, candidateEpoch uint64, options EpochBumpOptions) (*config.ChangeFeedInfo, error)
 	// PauseChangefeed persists the pause status to db for a changefeed
 	PauseChangefeed(ctx context.Context, id common.ChangeFeedID) error
 	// DeleteChangefeed removes all related info of a changefeed from db
 	DeleteChangefeed(ctx context.Context, id common.ChangeFeedID) error
 	// SetChangefeedProgress persists the operation progress status to db for a changefeed
 	SetChangefeedProgress(ctx context.Context, id common.ChangeFeedID, progress config.Progress) error
-	// ResumeChangefeed persists the resumed status to db for a changefeed and returns the resumed info.
-	ResumeChangefeed(ctx context.Context, id common.ChangeFeedID, newCheckpointTs uint64) (*config.ChangeFeedInfo, error)
```
3AceShowHand, line 40: Why remove this method ?

coordinator/changefeed/etcd_backend.go
```diff
@@ -207,6 +208,123 @@ func (b *EtcdBackend) UpdateChangefeed(ctx context.Context, info *config.ChangeF
 	return nil
 }
 
+// BumpChangefeedEpoch atomically persists a strictly newer ownership epoch.
```
3AceShowHand, line 210: It looks the resume chanegefeed method is replaced by this one? Looks not ideal, bump changefeed epoch should be a internal operation of the maintainer.

pkg/pdutil/utils.go
hongyunyan: Thanks, I agree that the non-PD part should not live in pdutil. I moved AdvanceChangefeedEpoch to pkg/common because it only compares persisted/candidate epochs and has no PD dependency. GenerateChangefeedEpoch stays in pdutil since it is specifically backed by PD TSO generation.

pkg/pdutil/utils.go
hongyunyan: It should not happen in normal operation because the candidate epoch is based on PD TSO/local timestamp. I kept the guard as a defensive invariant check because wrapping MaxUint64 back to 0 would make an old owner look newer or compatible again. The updated comment now calls this out as defensive rather than expected behavior.

coordinator/changefeed/changefeed.go
hongyunyan: For the nil info invariant, I moved the check to NewChangefeed/SetInfo as suggested. I kept the MarshalWithTruncation error handling here because this method serializes the current ChangeFeedInfo into the AddMaintainerRequest wire payload every time it sends the message. If serialization fails at this use site, continuing would send an invalid maintainer config, so keeping the local check makes the failure boundary explicit.

coordinator/changefeed/etcd_backend.go
hongyunyan: Agreed that the resume path should not be expressed as a raw bump at the controller call site. I restored Backend.ResumeChangefeed and made it own the resume-specific transition: StateNormal, ProgressNone, checkpoint update, error clear, and epoch bump in one backend operation. I kept BumpChangefeedEpoch as a lower-level backend boundary because other coordinator-driven ownership changes, such as add/move/retry, also must persist a fresh epoch before creating or moving a maintainer owner. So resume is semantic again, while the bump helper remains the shared metadata fence for non-resume owner-change paths.

3AceShowHand (approved):

asddongmen (approved):

lidezhu (approved):

Issue #5430, opened 2026-06-17
Optimize eventService scan path for heavy workload


asddongmen:
Type: Enhancement

Summary:

The eventService scan path can spend substantial CPU decoding and assembling DML rows that are later filtered out, especially in delete-heavy workloads where the changefeed ignores delete events. This issue tracks two focused improvements to reduce scan latency and resolved-ts lag under such workloads.

Plan:

1. Add a pre-decode DML filter fast path.
   - Detect filter decisions that only depend on startTs, table, and DML event type.
   - Skip mounter decode and DMLEvent append for rows that are definitively ignored, starting with ignored delete rows.
   - Add unit tests and workload validation for ignored delete cases.

2. Parallelize mounter work for rows that still require decoding.
   - Decode remaining rows with bounded worker parallelism.
   - Preserve original event order when emitting mounted rows.
   - Validate throughput, scan duration, CPU, heap, and sent resolved-ts lag before and after the change.

PR #5474, opened 2026-06-22
eventService: Optimize DML processing performance II 

asddongmen:
Issue Number: ref #5430 

Delete-heavy changefeeds that configure `ignore-event = ["delete"]` still decoded
ignored delete rows and appended them into `DMLEvent` before filtering, adding
unnecessary eventService scanner CPU during historical catch-up.

eventScanner now checks type-only DML filters before row decode for insert/delete
events and caches the result within each transaction. Ignored deletes bypass
`DecodeToChunk` and `DMLEvent.AppendRow`. Non-ignored deletes still use the
normal decode, append, and send path.

#### Tests

- Unit test
- Manual test on 140M row of TTL Deleted row, scan window off:

| Scenario | Master avg | Current avg | Result |
| --- | ---: | ---: | --- |
| Ignore delete | `338.3s` | `229.0s` | **`32.3%` faster 🔼** |
| No-ignore delete | `410.3s` | `400.7s` | neutral  |

No compatibility break expected. No-ignore delete keeps the normal path.

No. This is an internal eventService optimization.

### Release note

```release-note
Improve performance for changefeeds that ignore delete events.
```

## Summary by CodeRabbit

* **New Features**
  * Improved changefeed DML filtering with per-transaction caching and earlier “ignore by event type” decisions, reducing unnecessary work (including during update split handling).

* **Tests**
  * Added benchmarks covering DELETE-ignore behavior and comparing fast-path vs wrapped filtering.
  * Expanded unit/integration coverage to verify that ignored Insert/Delete/Update events skip decoding and produce the correct downstream DML and cache behavior.


pkg/filter/filter.go
```diff
@@ -47,6 +47,9 @@ const (
 type Filter interface {
 	// ShouldIgnoreDML returns true if the DML event should not be handled.
 	ShouldIgnoreDML(dmlType common.RowType, preRow, row chunk.Row, tableInfo *common.TableInfo, startTs uint64, ctx DMLFilterContext) (bool, error)
+	// ShouldIgnoreDMLByEventType returns true only when filters that do not need
+	// decoded row values can determine the DML should be ignored.
+	ShouldIgnoreDMLByEventType(dmlType common.RowType, tableInfo *common.TableInfo, startTs uint64) (bool, error)
```
lidezhu, line 52: Maybe `ShouldIgnoreDMLBeforeDecode` is a better name?

lidezhu (approved):
LGTM with minor comments

pkg/filter/filter.go
asddongmen: I prefer current name because I think filter shouldn't care if an event is decoded or not.

hongyunyan (approved):

wk989898 (approved):

asddongmen:
license/cla

asddongmen:
/license/cla

PR #5511, opened 2026-06-26
eventservice: split large scanned transactions

### LINKED ISSUE #5596: eventservice: EventBroker can OOM when syncing large transactions
Image

                      Image 1. CDC replicate 100 x 1GB insert transaction. 

When TiCDC syncs a very large upstream transaction, CDC can run out of memory.

The current EventBroker scan path may scan a whole transaction from eventstore and keep the scanned DML events in memory before sending them downstream. For very large transactions, this can build a large in-memory batch and cause OOM.

This is addressed by https://github.com/pingcap/ticdc/pull/5511.

asddongmen:
Issue Number: close #5596

EventBroker scans a whole large transaction, which can OOM CDC.

This PR adds row-level EventStore resume positions so `transaction-atomicity=none` emits bounded fragments without advancing resolved-ts early. Deferred insert halves of unique-key updates are spilled to disk and drained or cleaned across reset/remove.

#### TiUP E2E comparison

Runs used one CDC limited to 32 GiB, upstream/downstream TiDB, and the same workload.

| Workload | This PR | `master` |
| --- | --- | --- |
| 100 consecutive 1 GiB transactions | Synced without OOM | Repeated OOM/restart; could not finish syncing |
| One realtime 20 GiB transaction | OOMed once during ingestion; after restart, replayed as an incremental scan; checksum 33/33 | Never reached it: six earlier OOMs; no checksum |

The 20 GiB OOM is caused mainly by realtime logpuller `txnMatcher` PREWRITE caching and overlapping ingestion buffers, before EventBroker scan splitting applies. [asddongmen/tigate#2](https://github.com/asddongmen/tigate/pull/2) fixes this matcher pressure and passed a separate 20 GiB no-OOM test; whether to merge it will be evaluated separately.

#### Tests
- Unit test
- Integration test
- Manual test (results above)

The integration-test suite covers all changed workflows: row resume, interruption, unique-key spill/drain, CMEK spill encryption, dispatcher reset, and end-to-end consistency. Focused unit tests cover edge cases.

No compatibility break is expected. Splitting applies only when transaction atomicity permits it; unique-key updates may add local spill I/O.

An internal design document is included; no user-facing update is required.

### Release note
```release-note
Support splitting large transactions during EventBroker scanning to reduce OOM risk.
```


pkg/common/table_span.go
```diff
@@ -26,11 +26,17 @@ type DataRange struct {
 	CommitTsStart uint64
 	CommitTsEnd   uint64
 
+	// RowLevelScanPosition is an opaque eventstore-owned token used to resume
+	// scanning after a specific row inside the scan window.
+	RowLevelScanPosition ScanPosition
```
lidezhu: Non-blocking: I wonder whether `RowLevelScanPosition` and `LastScannedTxnStartTs` belong in `DataRange`.

`DataRange` conceptually describes the table span and commit-ts range, while these two fields describe where scanning should resume inside that range. Mixing them also makes operations such as `DataRange.Merge` ambiguous, since there is no obvious way to merge two different resume positions.

A cleaner model could be:
```
type ScanCursor struct {
    TxnStartTs uint64
    Position   ScanPosition
}

type ScanRequest struct {
    Range  DataRange
    Cursor ScanCursor
}
```

This would keep `DataRange` as a pure range and group the transaction-level and row-level resume states in one place. `TxnStartTs` could still be retained when `Position` is present, since EventService uses it to track an unfinished transaction while EventStore uses `Position` for precise row-level resume.

This is not a major issue and should not block this PR. The current design is functional, and we can clean up the model in a separate follow-up PR.

pkg/eventservice/dispatcher_stat.go
```diff
@@ -108,6 +108,13 @@ type dispatcherStat struct {
 	// These two values are used to construct the scan range for the next scan task.
 	lastScannedCommitTs atomic.Uint64
 	lastScannedStartTs  atomic.Uint64
+	lastScannedPosition atomic.Value
```
lidezhu: `lastScannedCommitTs`, `lastScannedStartTs`, and `lastScannedPosition` together represent one logical scan cursor, but they are currently stored and loaded independently.
It is better to store the complete progress as one immutable value instead.
```
type dispatcherScanProgress struct {
    commitTs uint64
    startTs  uint64
    position common.ScanPosition
}

scanProgress atomic.Pointer[dispatcherScanProgress]
```

pkg/eventservice/event_scanner.go
```diff
@@ -116,28 +146,42 @@ func (s *eventScanner) scan(
 	dispatcherStat *dispatcherStat,
 	dataRange common.DataRange,
 	limit scanLimit,
-) (int64, []event.Event, bool, error) {
+) (int64, []event.Event, scanProgress, bool, error) {
 	// Initialize scan session
 	sess := newSession(ctx, dispatcherStat, dataRange, limit)
 	defer sess.recordMetrics()
 
+	if state := dispatcherStat.getLargeTxnState(); state != nil &&
```
lidezhu: The atomic and split-transaction modes share the same event iteration and DDL/DML merge flow. Their main difference is how a transaction is started, appended, finished, and resumed. A strategy interface could keep these differences out of the common scanner loop.

```
  type txnScanStrategy interface {
      // resumePending resumes work left by a previous scan, such as draining
      // spilled inserts. handled is true when no EventStore scan is needed.
      resumePending(session *scanSession) (handled, interrupted bool, err error)

      // beforeRow handles mode-specific state before processing the next row.
      beforeRow(ctx *txnScanContext, row *common.RawKVEntry) (interrupted bool, err error)

      // startTxn starts a transaction with the atomicity behavior of this strategy.
      startTxn(ctx *txnScanContext, startTs, commitTs uint64, tableInfo *common.TableInfo) error

      // appendRow appends a row and may interrupt inside a splittable transaction.
      appendRow(ctx *txnScanContext, row *common.RawKVEntry, position common.ScanPosition) (interrupted bool, err error)

      // finishTxn commits the current transaction or transitions it to the next phase.
      finishTxn(ctx *txnScanContext, nextCommitTs uint64, tableInfoUpdateTs uint64, tableDeleted bool) (interrupted bool, err error)

      // cleanup releases mode-specific state after a scan failure.
      cleanup()
  }
```
`atomicTxnScanStrategy` would only interrupt at transaction boundaries, while `splitTxnScanStrategy` would own row-level interruption, spill, and drain.

This keeps one common scan loop while moving `largeTxnState` and spill-specific logic out of `eventScanner`.

lidezhu:
One thing to note: may be we need respect CMEK for the spill file?

wk989898 (approved):

lidezhu (approved):

asddongmen:
> One thing to note: may be we need respect CMEK for the spill file?

Yes, thank you. I have added that part.

PR #5547, opened 2026-06-30
eventservice: advance scan window for pending syncpoints

### LINKED ISSUE #5546: eventservice: scan window may stuck wheren there are pending syncpoints
## Problem

PR #5542 found that event service scan progress can stall when the global scan window is pinned by lagging dispatchers while another dispatcher has a pending syncpoint barrier in its current resolved range.

The existing local scan-window advance path handles pending DDL barriers, but syncpoint barriers need the same treatment so the dispatcher can make forward progress and eventually emit the syncpoint in order.

## Expected change

Split only the eventBroker change from #5542 onto a clean branch from `master`:

- detect pending syncpoint events before the uncapped end ts;
- locally advance the scan range when the global scan window makes the range empty;
- ensure the local range crosses `nextSyncPointTs` when needed;
- keep the behavior covered by focused `pkg/eventservice` unit tests.

Source PR: https://github.com/pingcap/ticdc/pull/5542

asddongmen:

Issue Number: close #5546

This PR splits the eventBroker runtime fix from #5542.

When the event service scan window is pinned by another lagging dispatcher, a dispatcher with a pending syncpoint barrier can get an empty scan range and fail to advance far enough to emit the syncpoint. The existing local scan-window advance path handled pending DDL barriers only.

- Detect pending syncpoint events before the uncapped scan end ts.
- Reuse the local scan-window advance path when either a DDL or syncpoint barrier is pending.
- Ensure the local range crosses `nextSyncPointTs` when the syncpoint would otherwise remain outside the bounded step.
- Add focused unit coverage for the pending syncpoint case.

Source PR: #5542

#### Tests

- Unit test

Commands run:
- `git diff --check`
- `PATH=/usr/local/go/bin:$PATH GOTOOLCHAIN=auto go test ./pkg/eventservice -run TestGetScanTaskDataRangeEmptyAfterCappingWithPendingSyncPointCrossesSyncPoint -count=1`
- `PATH=/usr/local/go/bin:$PATH GOTOOLCHAIN=auto make unit_test_pkg PKG=./pkg/eventservice/...`

No. The change only expands an existing local scan-window advance path to cover pending syncpoint barriers, matching the DDL barrier behavior.

No.

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scan-range handling so pending barrier events are respected when advancing work windows.
  * Prevented scan advancement from skipping over sync-point barriers, helping preserve correct event ordering.
  * Updated logging to reflect pending barrier events more accurately.
* **Tests**
  * Added coverage for scan-range behavior when a pending sync-point crosses the current scan window.


pkg/eventservice/event_broker.go
```diff
@@ -448,24 +450,32 @@ func (c *eventBroker) getScanTaskDataRange(task scanTask) (bool, common.DataRang
...
+		// For a table with pending ddl or syncpoint in current range, use a local bounded step
+		// to keep this dispatcher making forward progress, so barrier coverage can eventually complete.
 		interval := time.Duration(task.changefeedStat.scanInterval.Load())
 		if interval <= 0 {
 			interval = defaultScanInterval
 		}
 		localScanMaxTs := oracle.GoTimeToTS(oracle.GetTimeFromTS(dataRange.CommitTsStart).Add(interval))
+		if hasPendingSyncPointEventInCurrentRange && nextSyncPointTs >= dataRange.CommitTsStart &&
+			localScanMaxTs <= nextSyncPointTs {
+			localScanMaxTs = nextSyncPointTs + 1
```
wk989898, line 465: What will happen if localScanMaxTs Ts is less than nextSyncPointTs?

lidezhu (approved):

pkg/eventservice/event_broker.go
asddongmen: It may cause changefeed getting stuck, since syncpoint can not advance.

wk989898 (approved):

PR #5595, opened 2026-07-07
sink: add column selector for storage sink

### LINKED ISSUE #5592: support column selector in storage sink

No.

enable column selector in storage sink

_No response_

_No response_

wk989898:

Issue Number: close #5592

The column selector logic was generalized from a Kafka-only path into a shared
  column_selector flow that storage sink can also use.

It does three main things:

  - Adds column selector support to storage sink by carrying the selector on RowEvent before txn encoding.
  - Refactors canal/csv/kafka/storage encoding paths to share selector and callback handling instead of using sink-specific setters.
  - Adds validation so storage changefeeds cannot filter out all primary/unique key columns, preventing unsafe output for consumers.

#### About callback
  Storage keeps PostEnqueue and PostFlush as separate lifecycle events. PostEnqueue remains attached to the storage task and is triggered by the buffer/spool enqueue path. PostFlush is attached to the encoded message callback and runs after the encoded storage message is flushed.

For storage, the callback is intentionally different from Kafka.

  Kafka produces row-level messages. It uses a row callback counter: each row callback increments the count, and PostFlush runs only after all rows in the transaction have been flushed.

  Storage transaction encoders build a batch/transaction message. The encoder attaches only the last row callback to the built batch message, so the callback is triggered once per encoded storage message, not once per row. If storage used kafka's row-counting callback for a multi-row transaction, the counter would never reach the row count and PostFlush could hang forever.

  That is why storage directly passes event.PostFlush as the row callback. The encoded storage batch message calls it once when the message is flushed, which matches storage's actual callback granularity.

#### Tests 

- Unit test
- Integration test

### Release note 

```release-note
Support column selector for storage sink
```

## Summary
* **New Features**
  * Added end-to-end column-selector support for storage and MQ pipelines (cloud storage, Kafka, Pulsar, and CSV/JSON), including per-table filtering that affects both encoded output and CSV headers/values.
  * Storage consumer now applies column selectors when decoding CSV DML files.
* **Bug Fixes**
  * Improved DML task callback flow to ensure post-flush and post-enqueue hooks run reliably and in the expected order.
  * Updated encoders to consistently use row-level event data for txn message creation.
* **Tests**
  * Expanded unit and integration coverage for column selection, CSV decoding/encoding, and callback lifecycle behavior.


downstreamadapter/sink/cloudstorage/task.go
```diff
@@ -51,16 +51,14 @@ type task struct {
...
-	// Sink.AddDMLEvent, so snapshot callbacks here and release the large event
-	// object after encoding.
 	return &task{
 		kind:           taskKindDML,
-		event:          event,
-		callbacks:      newTxnCallbacks(event),
+		postEnqueue:    event.PostEnqueue,
 		tableInfo:      event.TableInfo,
 		versionedTable: version,
+		rowEvents:      helper.NewRowEvents(event, selector, event.PostFlush),
```
asddongmen: Could we keep the lightweight callback copy here? Passing `event.PostFlush` / `event.PostEnqueue` as method values keeps the whole `DMLEvent` alive until storage flush or quota release. The previous `txnCallbacks` avoided retaining large Rows after encoding.

downstreamadapter/sink/cloudstorage/sink.go
```diff
@@ -86,6 +87,9 @@ func Verify(ctx context.Context, changefeedID common.ChangeFeedID, sinkURI *url.
 	if err != nil {
 		return err
 	}
+	if _, err = columnselector.New(sinkConfig); err != nil {
```
asddongmen, line 90: This only validates the selector syntax for storage. The table-level `selectors.VerifyTables` check is still gated by `config.IsMQScheme` in `verifyTable4MQ`, so storage can accept a selector that filters out PK/UK columns. It may make delete/update replay incorrect.

asddongmen (commented):
Two comments.

asddongmen (approved):

3AceShowHand (approved):

PR #5610, opened 2026-07-09
logpuller: fix potential shutdown hang in paused region event push

### LINKED ISSUE #5608: logpuller: paused region event push can block graceful shutdown
### What did you do?

The region event push path can block graceful shutdown when event pushing is paused.

Current behavior:

  - when memory control pauses pushing, the push path waits on `sync.Cond`
  - the waiter checks cancellation before entering `cond.Wait()`
  - once it is already blocked in `cond.Wait()`, context cancellation does not wake it up
  - the waiter is only released by a later `ResumeArea` or another explicit `Broadcast()`

Why this can hang shutdown:

  1. `subscriptionClient.Run(ctx)` uses an errgroup and waits for all worker goroutines to exit.
  2. region request workers are part of that same errgroup.
  3. workers can call the region event push path while cleaning up region state, handling region events, or flushing resolved-ts batches.
  4. if pushing is paused at that time, a worker goroutine can block in `cond.Wait()`.
  5. on shutdown, the top-level context is canceled first, and the process waits for `Run(ctx)` to return.
  6. `Close()` is only called after `Run(ctx)` returns.
  7. if the blocked worker is still waiting for a pause/resume signal, it can keep `Run(ctx)` from finishing.
  8. but the extra wake-up that would release it may only happen during `Close()`.

This creates a shutdown dependency cycle:

  - shutdown cancels `Run(ctx)`
  - `Run(ctx)` waits for workers to exit
  - a worker is blocked in paused push wait
  - the wake-up for that wait may only happen in later cleanup

As a result, graceful shutdown can be delayed or hang until timeout.

_No response_

### What did you see instead?

None

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```

lidezhu:

Issue Number: close #5608 

This pull request improves the reliability and shutdown behavior of the region event sink by making the paused push path interruptible. By introducing a 'stopped' atomic flag and refining the synchronization logic, the system can now gracefully handle shutdown signals even when event processing is temporarily paused due to flow control, preventing potential goroutine leaks or hangs.

### Highlights

* **Interruptible Event Pushing**: Introduced a 'stopped' state to the region event sink, allowing blocked push operations to exit immediately during shutdown instead of waiting for a resume signal.
* **Refactored Synchronization**: Replaced context-based waiting in the 'Push' method with a more robust condition variable and atomic flag approach, simplifying the lifecycle management of the sink.
* **Encapsulated State Management**: Consolidated pause, resume, and stop logic into dedicated helper methods to improve code readability and maintainability.

#### Tests 

- Unit test
- Integration test
- No code

### Release note 

```release-note

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown handling for paused event delivery, preventing hangs when stopping during a paused state.
  * Ensured `Push` operations unblock promptly on run cancellation, improving pause/resume reliability.
* **Tests**
  * Added coverage to verify cancellation unblocks paused event pushes and that the sink exits cleanly.
  * Centralized test sink setup and updated related tests/bench scaffolding to use consistent sink behavior.

### DIFF
--- a/logservice/logpuller/region_event_sink.go
+++ b/logservice/logpuller/region_event_sink.go
@@ -26,16 +26,19 @@ import (
 
 // regionEventSink delivers region events to dynstream and owns push-side flow control.
 type regionEventSink struct {
-	ctx context.Context
-	ds  dynstream.DynamicStream[int, SubscriptionID, regionEvent, *subscribedSpan, *regionEventHandler]
-	// the following three fields are used to manage feedback from ds and notify other goroutines
-	mu     sync.Mutex
-	cond   *sync.Cond
-	paused atomic.Bool
+	ds dynstream.DynamicStream[int, SubscriptionID, regionEvent, *subscribedSpan, *regionEventHandler]
+	// the following fields are used to coordinate pause/resume feedback with event producers.
+	mu       sync.Mutex
+	resumeCh chan struct{}
+	stopCh   chan struct{}
+	stopOnce sync.Once
+	paused   atomic.Bool
 }
 
-func newRegionEventSink(ctx context.Context, failureHandler *regionFailureHandler) *regionEventSink {
-	sink := &regionEventSink{ctx: ctx}
+func newRegionEventSink(_ context.Context, failureHandler *regionFailureHandler) *regionEventSink {
+	sink := &regionEventSink{
+		stopCh: make(chan struct{}),
+	}
 
 	option := dynstream.NewOption()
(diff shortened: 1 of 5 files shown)


lidezhu:
@coderabbitai review

hongyunyan (approved):

asddongmen (approved):

PR #5718, opened 2026-07-24
kafka: share one claimcheck instance across encoders

### LINKED ISSUE #5719: kafka: avoid creating a ClaimCheck storage client for every encoder
## Problem

When Kafka claim-check handling is enabled, each event encoder creates its own `ClaimCheck` instance. An encoder group contains multiple encoders, and additional standalone encoders may also be created, so one Kafka sink can open multiple external storage clients for the same claim-check URI.

The number of storage clients grows with encoder concurrency even though all encoders belong to the same sink and use the same configuration. This wastes client-side resources and makes ownership and cleanup of the external storage instances unclear.

## Expected behavior

A Kafka sink should:

- create one `ClaimCheck` instance after the encoder configuration is built;
- share that instance with all encoders owned by the sink;
- keep claim-check message encoding and configuration behavior unchanged;
- close the underlying external storage when the Kafka sink is closed.

3AceShowHand:

When Kafka claim-check handling is enabled, every event encoder creates its own `ClaimCheck` and external storage client. Because an encoder group contains multiple encoders, the number of storage clients grows with encoder concurrency even though they all use the same configuration and storage URI.

Issue Number: close #5719

- Create one `ClaimCheck` at the Kafka sink component level.
- Inject the same instance into the encoder group and the standalone encoder.
- Close the underlying claim-check storage when the Kafka sink component is closed.
- Remove encoder-owned claim-check metric cleanup because ownership now belongs to the Kafka sink component.
- Update codec and claim-check tests to use the shared instance and verify concurrent writes and storage cleanup.

Pulsar does not enable claim-check handling and continues to pass no `ClaimCheck` to its encoders.

#### Tests

- Unit test

  ```text
  go test --tags=intest ./pkg/sink/kafka/claimcheck ./pkg/sink/codec/open ./pkg/sink/codec/canal ./pkg/sink/codec/simple ./downstreamadapter/sink/kafka ./downstreamadapter/sink/pulsar -count=1
  go test -race --tags=intest ./pkg/sink/kafka/claimcheck -count=1
  ```

It reduces external storage client and connection usage when Kafka claim-check is enabled. It changes exported Go encoder constructor signatures to accept a sink-owned `ClaimCheck`, but does not change sink configuration, encoded message formats, or runtime protocol compatibility.

No user-facing documentation or monitoring changes are required because configuration and behavior remain unchanged.

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  * Improved handling of oversized Kafka messages through shared claim-check storage across supported encoders.
  * Added safer resource lifecycle management for large-message storage.

* **Bug Fixes**
  * Improved concurrent large-message writes and cleanup behavior.
  * Prevented resource leaks when sink or encoder initialization fails.

* **Tests**
  * Expanded coverage for concurrent writes, storage closure, and large-message encoding.


lidezhu (approved):

downstreamadapter/sink/kafka/sink.go
```diff
@@ -95,6 +96,12 @@ func Verify(ctx context.Context, changefeedID commonType.ChangeFeedID, uri *url.
 		return errors.Trace(err)
 	}
 
+	claimCheck, err := claimcheck.New(ctx, encoderConfig.LargeMessageHandle, changefeedID)
+	if err != nil {
+		return err
+	}
+	defer claimCheck.Close()
```
wk989898, line 103:
```suggestion
	if claimCheck!= nil {
		defer claimCheck.Close()
	}
```

asddongmen (approved):

PR #5758, opened 2026-07-26
logpuller: keep bootstrap requests flow-controlled

### LINKED ISSUE #5748: logpuller: high-priority bootstrap bypasses region flow control
### What did you do?

Created a changefeed on three TiCDC captures for 32 tables with 6,248 Regions
per table (199,936 record Regions in total). The changefeed used a fresh
start-ts, so the initial scans were inside the default scan-priority lag
threshold.

The bootstrap was tested twice on exact master
`d480b05fa79b596204b2e881c7184a5f2902ffa2`. The second attempt cold-restarted
all three TiCDC captures before creating the changefeed.

Other controlled settings:

- one TiKV store;
- TiDB `min-ts-interval = 200ms`;
- MySQL dry-run sink with 50ms simulated latency;
- no DML workload during bootstrap;
- default TiCDC `pending-region-request-queue-size = 32`;
- default TiKV `cdc.incremental-scan-concurrency-limit = 10000`.

The initial scans should remain bounded by TiCDC's pending-Region flow control,
finish registering all Regions, and let resolved-ts/checkpoint-ts advance past
the changefeed start-ts.

Historical master `369674c75814928e5a3f1122fc7520c370033571`
completed the same 199,936-Region bootstrap in 206.9 seconds.

### What did you see instead?

Both exact-master attempts failed to advance past start-ts within 20 minutes.
The controlled second attempt ran from
`2026-07-25 04:27:36.633` to `04:47:12.818` CST.

Metrics in that window:

| Metric | Observed value |
| --- | ---: |
| `ticdc_subscription_client_requested_region_count{state="pending"}` | aggregate peak 198,320 |
| `increase(ticdc_kvclient_event_feed_error_count{type="KvIsBusy"}[19m])` | 40.69m / 35.93m / 41.09m on the three captures; about 119.6m total |
| Other `ticdc_kvclient_event_feed_error_count` types | 0 |
| TiKV unresolved CDC Regions | exactly 10,000 |
| TiKV captured CDC Regions at the end | about 15,507 |
| LogPuller Region matchers | about 200,397 |
| EventStore resolved input | 0 |
| Successful Region scan progress during the storm | about 4-5 Regions/s |

For comparison, two successful builds with the same topology had zero
`KvIsBusy`, reached about 200,099 captured Regions with zero unresolved
Regions, and peaked at only 89 and 96 aggregate pending Region requests.

At default INFO level, TiCDC records `cdc region error` only at DEBUG, so the
`ServerIsBusy` reason is not printed in TiCDC logs. The error counter above is
the direct TiCDC signal. TiKV's 300 MiB log files rotated roughly every
12 seconds during the retry storm, dominated by per-Region incremental-scan
messages such as:

```text
[2026/07/25 04:44:52.763 +08:00] [INFO] [initializer.rs:542]
["cdc has scanned all incremental scan locks"] ... [region_id=562097]
[2026/07/25 04:44:52.764 +08:00] [INFO] [initializer.rs:388]
["cdc async incremental scan finished"] ... [region_id=562097]
```

TiKV v8.5.5 rejects registrations after the configured/default 10,000
incremental-scan tasks with:

```text
ServerIsBusy("too many pending incremental scans")
```

### Suspected cause

The behavior is introduced by scan-priority commit `d8b039fd5` / PR #5480:

1. A recent start-ts makes initial subscription scans high priority.
2. `subscription_client.go` derives
   `force := regionTask.Priority() <= forcedPriorityBase`.
3. `requestCache.add(..., force=true)` bypasses the
   `pendingCount < maxPendingCount` flow-control check.
4. `region_failure_handler.go` preserves the inherited high priority for
   `ServerIsBusy` and `Congested`.
5. The rejected task is immediately requeued as high priority, bypasses flow
   control again, and has no backoff or jitter.

The deployed TiKV v8.5.5 uses a kvproto version that does not consume the new
`ChangeDataRequest.scan_priority` field. Therefore the intended TiKV-side
priority behavior is unavailable, while TiCDC's local `force` and retry
behavior still take effect.

The full-span initialization change in `dd07f601b` makes the frozen start-ts
visible and correct, but does not create the request storm.

### Suggested fix

Priority and local admission bypass should be separate decisions:

- initial subscription scans may carry high scan priority to TiKV, but should
  still honor the per-store pending-Region limit;
- `ServerIsBusy` and `Congested` retries should be demoted behind normal
  flow control and use bounded backoff with jitter;
- `force=true` should be reserved for a small, bounded class of recovery or
  control work, not all recent-start initial scans.

As a narrow hotfix, demoting `ServerIsBusy`/`Congested` retries to
`TaskLowPrior` should break the positive feedback loop. It is not a complete
fix by itself because the first high-priority bootstrap burst can still bypass
the local pending limit before TiKV returns the first busy response.

### Versions

- TiCDC: master `d480b05fa79b596204b2e881c7184a5f2902ffa2`
- TiDB: v8.5.5, commit `1fa258b833ff113883beeba40bc130be7ce66610`
- TiKV: v8.5.5, commit `a0234aad114258b60e0ec23c17486c79845299be`

asddongmen:

Issue Number: close #5748

Remote HIGH scan priority also enabled local forced admission. Bootstrap and
overload retries could bypass the per-worker limit, stall bootstrap, and affect
existing changefeeds.

- Separate remote scan priority from local queue/admission priority.
- Retry `ServerIsBusy`/`Congested` locally as LOW while preserving remote HIGH.
- Reserve slots before publishing and cap repair admission at `N+1`; stop/control
  behavior is unchanged.

#### Priority semantics

| Scenario | Local queue | Remote `ScanPriority` | Data admission |
|---|---|---|---|
| New changefeed, old start-ts bootstrap | LOW | LOW | Normal `N` |
| New changefeed, recent start-ts bootstrap | LOW | HIGH | Normal `N` |
| Existing caught-up changefeed, split/merge/NotLeader repair | HIGH | HIGH | Up to `N+1` |
| Existing caught-up changefeed, `ServerIsBusy`/`Congested` retry | LOW | HIGH | Normal `N` |

Examples:

- Split/merge `EpochNotMatch` creates HIGH local/remote repair Regions that may
  use the bounded repair slot.
- A recent bootstrap keeps remote HIGH, but remains local LOW and cannot use
  the repair slot.
- A busy retry remains remote HIGH for CSE, but uses normal local admission to
  avoid recreating the overload loop.

#### Tests

- [x] Unit test
- [x] Manual test

Three captures bootstrapped 199,936 Regions in 285.05s: pending peaked at 96
with zero `KvIsBusy`. The regressed build peaked at 198,320, produced about
119.6 million busy errors, and stalled beyond 20 minutes.

No compatibility change. Bootstrap concurrency is intentionally bounded.

No.

### Release note

```release-note
Fix a potential TiCDC bootstrap stall caused by unbounded Region requests when TiKV is overloaded.
```

### DIFF
--- a/logservice/logpuller/region_failure_handler.go
+++ b/logservice/logpuller/region_failure_handler.go
@@ -138,12 +138,12 @@ func (r *regionFailureHandler) handleError(ctx context.Context, errInfo regionEr
 		}
 		if innerErr.GetCongested() != nil {
 			metricKvCongestedCounter.Inc()
-			r.client.scheduleRegionRequest(ctx, errInfo.regionInfo, retryPriority)
+			r.client.scheduleRegionRequest(ctx, errInfo.regionInfo, TaskLowPrior)
 			return nil
 		}
 		if innerErr.GetServerIsBusy() != nil {
 			metricKvIsBusyCounter.Inc()
-			r.client.scheduleRegionRequest(ctx, errInfo.regionInfo, retryPriority)
+			r.client.scheduleRegionRequest(ctx, errInfo.regionInfo, TaskLowPrior)
 			return nil
 		}
 		if duplicated := innerErr.GetDuplicateRequest(); duplicated != nil {
--- a/logservice/logpuller/region_req_cache.go
+++ b/logservice/logpuller/region_req_cache.go
@@ -61,8 +61,9 @@ type requestCache struct {
 	}
 
 	// pendingCount is a flow control slot counter.
-	// A slot is acquired when a request is successfully enqueued into pendingQueue (see add),
+	// A slot is acquired before publishing a request into pendingQueue (see add),
 	// and is released when the request is finished/removed (resolve/markStopped/markDone/clear).
+	// add releases the slot itself if queue publication fails.
 	// pop and markSent don't change it. If markSent overwrites an existing request for the same region,
 	// it will release a slot for the replaced request to avoid leaking pendingCount.
 	pendingCount atomic.Int64
@@ -92,7 +93,8 @@ func newRequestCache(maxPendingCount int) *requestCache {
 }
 
 // add adds a new region request to the cache
-// It blocks if pendingCount >= maxPendingCount until there's space or ctx is cancelled
+// Normal data requests are limited to maxPendingCount. Forced data requests can
+// use one additional slot, while stop/control requests keep their existing bypass.
 func (c *requestCache) add(ctx context.Context, region regionInfo, force bool) (bool, error) {
 	start := time.Now()
 	ticker := time.NewTicker(addReqRetryInterval)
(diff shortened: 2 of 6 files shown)


asddongmen:
This PR becomes unnecessary because of https://github.com/pingcap/ticdc/pull/5619

PR #5779, opened 2026-07-27
coordinator: handle errors from regressed checkpoint status

### LINKED ISSUE #5709: Maintainer errors are dropped when checkpoint regresses, preventing automatic changefeed recovery
## What did you do?

We observed a production incident in the new-architecture TiCDC with a Kafka sink.

During a transient connectivity disturbance:

1. One capture lost its etcd session and restarted.
2. The capture membership change caused many table dispatchers of one changefeed to be rescheduled.
3. During rescheduling, the maintainer checkpoint regressed below the checkpoint already stored by the coordinator.
4. At the same time, a dispatcher reported a retryable `ErrKafkaAsyncSendMessage`.
5. The affected changefeed never entered warning/backoff and did not recover automatically. A manual changefeed restart was required.

All customer, cluster, changefeed, table, network, and DML identifiers below are masked.

### Sanitized timeline

The capture restart was caused by the TiCDC process fencing itself after its etcd session ended, rather than by OOM or a Kubernetes eviction:

```text
[2026/07/22 16:30:08.634 +00:00] [WARN] [etcd_watcher.go:70]
["session is disconnected"]
[error="[CDC:ErrEtcdSessionDone]etcd session is done"]

[2026/07/22 16:30:08.672 +00:00] [ERROR] [server.go:<line>]
["cdc server exits with error"]
[error="[CDC:ErrCaptureSuicide]capture suicide"]
```

The affected changefeed maintainer then observed the old capture leaving and a new capture joining:

```text
[2026/07/22 16:30:13.486 +00:00] [INFO] [maintainer.go:540]
["maintainer node changed"]
[changefeedID=default/<changefeed>]
[addedCount=0] [removedCount=1]
[removedNodes=["<old-capture-id>"]]

[2026/07/22 16:30:15.987 +00:00] [INFO] [maintainer.go:540]
["maintainer node changed"]
[changefeedID=default/<changefeed>]
[addedCount=1] [removedCount=0]
[addedNodes=["<new-capture-id>"]]
```

The incident queued more than 100 add-dispatcher operators for this changefeed. Around the same time, another capture reported a Kafka send timeout:

```text
[2026/07/22 16:30:15.758 +00:00] [ERROR] [maintainer.go:799]
["dispatcher report an error"]
[changefeedID=default/<changefeed>]
[sourceNode=<capture-id>]
[error="[CDC:ErrKafkaAsyncSendMessage]kafka async send message failed:
        keyspace=default, changefeed=<changefeed>, eventType=dml,
        dmlInfo=<redacted>;
        ErrorInfo:read tcp <ticdc-address>:<port>-><kafka-broker>:9092: i/o timeout"]
```

Metrics show that the maintainer checkpoint regressed during failover. Values are expressed relative to a redacted base physical timestamp `T`:

```text
16:30:07  maintainer checkpoint = T + 4.800s
16:30:22  maintainer checkpoint = T          # regressed by 4.800s
           owner checkpoint retained = T + 5.750s

Later:
           maintainer checkpoint = T + 1.800s
           owner checkpoint      = T + 5.750s
           difference            = 3.950s
```

After the regression, the same Kafka error was reported by the maintainer every five seconds. Log-site grouping found one original dispatcher-manager error followed by 289 maintainer reports. The last report was about 30 minutes after the first one.

There were no `backoff.go` entries for the affected changefeed. In the same incident, at least six other changefeeds with Kafka errors did reach:

```text
[ERROR] [backoff.go:157] ["changefeed maintainer report an error"]
[WARN]  [backoff.go:220] ["changefeed meets an error, will be stopped"]
```

Those changefeeds were automatically stopped/recreated and recovered. The affected changefeed remained in `normal` state while checkpoint lag grew linearly.

A manual changefeed restart closed all dispatcher managers and recreated the sink clients:

```text
[2026/07/22 17:00:18.709 +00:00] [INFO] [dispatcher_manager.go:889]
["closing event dispatcher manager"]
[changefeedID=default/<changefeed>]

[2026/07/22 17:00:41.164 +00:00] [INFO] [dispatcher_manager.go:306]
["dispatcher manager initialized"]
[changefeedID=default/<changefeed>]
[maintainerID=<new-maintainer-id>]
[startTs=<redacted>]
```

After the restart, checkpoint lag returned to about 3 seconds and no new Kafka async-send error was observed.

A retryable error in `MaintainerStatus.Err` should be handled even when the incoming checkpoint is lower than the coordinator's last stored checkpoint.

The coordinator should:

1. Keep the stored checkpoint monotonic.
2. Still process the error and transition the changefeed to warning/backoff.
3. Stop and recreate the changefeed automatically.

A safe checkpoint regression can occur during node failover because absent/scheduling spans participate in the maintainer's global checkpoint minimum. It should not disable error recovery.

## What did you see instead?

`Changefeed.UpdateStatus` gates both checkpoint storage **and error handling** on the incoming checkpoint being monotonic:

```go
if newStatus != nil && newStatus.CheckpointTs >= old.CheckpointTs {
    c.status.Store(newStatus)
    // ...
    return c.backoff.CheckStatus(newStatus)
}

return false, config.StateNormal, nil
```

When the maintainer checkpoint regresses, the whole status is ignored, including `newStatus.Err`. As long as subsequent reports remain below the coordinator checkpoint:

- `Backoff.CheckStatus` is never called.
- The changefeed never enters warning.
- Automatic stop/recreate is never triggered.
- The dispatcher manager keeps re-reporting its first terminal sink error until it is closed manually.

Relevant production-version code:

- [checkpoint gate in Changefeed.UpdateStatus](https://github.com/pingcap/ticdc/blob/e7e1f6142308f893acbd2aea8ee57af49a53e5fc/coordinator/changefeed/changefeed.go#L149-L173)
- [absent/scheduling spans constrain the maintainer checkpoint](https://github.com/pingcap/ticdc/blob/e7e1f6142308f893acbd2aea8ee57af49a53e5fc/maintainer/span/span_controller.go#L197-L208)
- [dispatcher manager resends the first error until close](https://github.com/pingcap/ticdc/blob/e7e1f6142308f893acbd2aea8ee57af49a53e5fc/downstreamadapter/dispatchermanager/dispatcher_manager.go#L501-L555)

The same lower-checkpoint early-return behavior is still present on the current `master` branch at the time of filing.

### Suggested fix and regression test

Decouple checkpoint monotonicity from error handling:

- Do not overwrite the stored checkpoint with a lower value.
- Do not discard `MaintainerStatus.Err`.
- Process the retryable error using the stored checkpoint (or otherwise route it to backoff independently).

Add a test covering:

```text
old stored checkpoint = 200
incoming checkpoint   = 150
incoming Err          = retryable sink error

expected:
- stored checkpoint remains 200
- state changes to warning
- retry/backoff starts
```

This is related by symptom to #3952, but the root cause is different. In #3952 the error reached warning/backoff and recovery failed later during maintainer re-add. In this incident the error is dropped before backoff starts.

## Versions of the cluster

Upstream TiDB cluster version:

```console
v8.5 series (exact binary version was not separately collected)
```

Upstream TiKV version:

```console
v8.5 series (exact binary version was not separately collected)
```

TiCDC version:

```console
Release Version: v8.5.7-release.1
Git Commit Hash: e7e1f6142308f893acbd2aea8ee57af49a53e5fc
```

wk989898:

Issue Number: close #5709

  Before this change, `Changefeed.UpdateStatus` ignored maintainer status reports whose `CheckpointTs` was lower than the checkpoint already stored in the coordinator. This kept checkpoint updates monotonic, but it also dropped `MaintainerStatus.Err` carried by those reports. As a result, retryable maintainer errors could be missed and the changefeed would not enter warning/backoff as expected.

  This change keeps the monotonic checkpoint rule while still processing reported errors. If a maintainer reports a lower checkpoint without errors, the status is ignored as before. If it reports a lower checkpoint with errors, TiCDC copies the status, replaces the copied `CheckpointTs` with the coordinator's current checkpoint, and sends the copied status through the existing fast-fail and backoff checks.

  With this behavior, maintainer errors can update the changefeed state and trigger retry/backoff, while the stored checkpoint never decreases and the original maintainer status object is not mutated.
#### Tests 

- Unit test
- Integration test

### Release note 

```release-note
 Fix an issue where TiCDC could ignore maintainer errors when the reported checkpoint regressed.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved changefeed status handling when no new status is provided.
  * Ensured retryable errors are processed correctly even when status checkpoints move backward.
  * Preserved accurate warning state, error details, and retry behavior during checkpoint regressions.

### DIFF
--- a/coordinator/changefeed/changefeed.go
+++ b/coordinator/changefeed/changefeed.go
@@ -161,12 +161,25 @@ func (c *Changefeed) ShouldRun() bool {
 // It returns false if the status is not changed
 // It returns the new state and error if the status is changed
 func (c *Changefeed) UpdateStatus(newStatus *heartbeatpb.MaintainerStatus) (bool, config.FeedState, *heartbeatpb.RunningError) {
+	if newStatus == nil {
+		return false, config.StateNormal, nil
+	}
+
 	old := c.status.Load()
 	failpoint.Inject("CoordinatorDontUpdateChangefeedCheckpoint", func() {
 		newStatus.CheckpointTs = old.CheckpointTs
 	})
 
-	if newStatus != nil && newStatus.CheckpointTs >= old.CheckpointTs {
+	if newStatus.CheckpointTs < old.CheckpointTs {
+		if len(newStatus.Err) == 0 {
+			return false, config.StateNormal, nil
+		}
+		statusWithMonotonicCheckpoint := *newStatus
+		statusWithMonotonicCheckpoint.CheckpointTs = old.CheckpointTs
+		newStatus = &statusWithMonotonicCheckpoint
+	}
+
+	if newStatus.CheckpointTs >= old.CheckpointTs {
 		c.status.Store(newStatus)
 
 		changed, state, err := c.backoff.checkFailedStatus(newStatus)
--- a/coordinator/changefeed/changefeed_test.go
+++ b/coordinator/changefeed/changefeed_test.go
@@ -110,6 +110,45 @@ func TestChangefeed_UpdateStatus(t *testing.T) {
 	require.Equal(t, newStatus, cf.GetStatus())
 }
 
+func TestChangefeed_UpdateStatusProcessesErrorsWhenCheckpointRegresses(t *testing.T) {
+	cfID := common.NewChangeFeedIDWithName("test", common.DefaultKeyspaceName)
+	info := &config.ChangeFeedInfo{
+		SinkURI: "kafka://127.0.0.1:9092",
+		State:   config.StateNormal,
+		Config:  config.GetDefaultReplicaConfig(),
+	}
+	cf := NewChangefeed(cfID, info, 200, true)
+
+	regressedStatus := &heartbeatpb.MaintainerStatus{CheckpointTs: 150}
+	updated, state, err := cf.UpdateStatus(regressedStatus)
+	require.False(t, updated)
+	require.Equal(t, config.StateNormal, state)
+	require.Nil(t, err)
+	require.Equal(t, uint64(200), cf.GetStatus().CheckpointTs)
+
+	retryableErr := &heartbeatpb.RunningError{
+		Node:    "node-1",
+		Code:    "CDC:ErrChangefeedRetryable",
+		Message: "retryable error",
+	}
+	regressedStatus = &heartbeatpb.MaintainerStatus{
+		CheckpointTs: 150,
+		Err:          []*heartbeatpb.RunningError{retryableErr},
+	}
+	updated, state, err = cf.UpdateStatus(regressedStatus)
+	require.True(t, updated)
+	require.Equal(t, config.StateWarning, state)
+	require.Same(t, retryableErr, err)
+	require.Equal(t, uint64(150), regressedStatus.CheckpointTs)
+	status := cf.GetStatus()
+	require.Equal(t, uint64(200), status.CheckpointTs)
+	require.Len(t, status.Err, 1)
+	require.Same(t, retryableErr, status.Err[0])
+	require.False(t, cf.ShouldRun())
+	require.True(t, cf.backoff.retrying.Load())
+	require.True(t, cf.backoff.isRestarting.Load())
+}
+
 func TestChangefeed_UpdateStatusFastFailWhenBootstrapDoneChanges(t *testing.T) {
 	cfID := common.NewChangeFeedIDWithName("test", common.DefaultKeyspaceName)
 	info := &config.ChangeFeedInfo{


asddongmen:
Will this casue some stale errors to be handled unintentionally?

wk989898:
> Will this casue some stale errors to be handled unintentionally?

 Not for stale maintainers / stale epochs. The new behavior can only run after the coordinator has already accepted the status as coming from the current maintainer node and current maintainer epoch.

lidezhu (approved):

asddongmen (approved):

Issue #5705, opened 2026-07-22
*: optimize resolved-ts and checkpoint propagation latency


asddongmen:
## Is your feature request related to a problem?

The TiCDC resolved-ts and checkpoint propagation path contains several avoidable periodic or next-notification waits. Even when upstream TiKV advances resolved-ts every 200 ms, batching and reporting intervals across LogPuller, EventService, DispatcherManager, and Maintainer can accumulate into noticeably higher changefeed latency.

Some dispatchers can also wait for the next EventStore notification when a scan is already running or when SchemaStore temporarily caps the scan frontier.

## Describe the feature you would like

Provide an opt-in low-latency performance mode while keeping the current throughput-oriented behavior as the default.

The low-latency path should reduce unnecessary waits in resolved-ts advancement, dispatcher and Maintainer watermark reporting, checkpoint calculation, and scan continuation or schema-blocked recovery. Work must remain bounded and coalesced, with explicit CPU and control-message trade-offs.

Validate the design at large table and Region counts, covering steady-state latency, bootstrap, queue saturation, resource usage, and throughput-mode compatibility.

Prototype and initial scale-test results: https://github.com/asddongmen/tigate/pull/4

PR #5826, opened 2026-07-31
*: add low-latency watermark propagation

asddongmen:

Issue Number: ref #5705

This is PR2, the non-EventService subset split from PR1 [#5749](https://github.com/pingcap/ticdc/pull/5749) so the low-latency propagation changes can be reviewed independently.

PR2 intentionally contains no `pkg/eventservice` code or EventService metric changes; those changes remain in PR1.

- Add opt-in `performance-mode = "low-latency"` while keeping throughput mode as the default and rejecting unknown values.
- Advance LogPuller resolved-ts without timestamp batching when low-latency mode is enabled.
- Start dispatcher-manager heartbeats immediately and reduce their interval from 200 ms to 50 ms in low-latency mode.
- Wake maintainer checkpoint calculation as soon as an accepted dispatcher watermark changes in low-latency mode.
- Report maintainer watermarks every 50 ms and promptly mark changed watermarks for coordinator reporting in low-latency mode.
- Refresh log-coordinator resolved-ts metrics after a complete node reporting round and report the maximum per-node lag.

#### Tests

- Unit test (`GOFLAGS=-mod=mod make unit_test_pkg PKG='<five owning package scopes>'`; 459 tests passed).
- Manual test (PR1 contains the three-capture, 10-minute table and Region benchmark matrices; this split did not rerun them).

Throughput mode remains the default with unchanged behavior, while low-latency mode intentionally increases heartbeat frequency. Enable it only after all TiCDC nodes are upgraded, and remove the option before downgrading.

Yes, the performance mode and its resource trade-off need user documentation, which remains tracked by PR1 and is not duplicated in this split.

### Release note

```release-note
Add an opt-in low-latency mode for faster LogPuller, dispatcher heartbeat, maintainer checkpoint, and log-coordinator metric propagation.
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable performance modes: **throughput** (default) and **low-latency**.
  * Low-latency mode provides faster heartbeats, immediate resolved-timestamp advancement, and more responsive checkpoint updates.
* **Improvements**
  * Improved changefeed metrics and status reporting, including more timely updates as nodes report progress.
  * Enhanced watermark and resolved-timestamp tracking to better reflect current system state.
* **Validation**
  * Added configuration validation to reject unsupported performance mode values.

### DIFF
--- a/downstreamadapter/dispatchermanager/task.go
+++ b/downstreamadapter/dispatchermanager/task.go
@@ -24,8 +24,15 @@ import (
 	"github.com/pingcap/ticdc/heartbeatpb"
 	"github.com/pingcap/ticdc/pkg/common"
 	appcontext "github.com/pingcap/ticdc/pkg/common/context"
+	"github.com/pingcap/ticdc/pkg/config"
 	"github.com/pingcap/ticdc/utils/threadpool"
 	"go.uber.org/zap"
+)
+
+const (
+	defaultHeartbeatInterval     = 200 * time.Millisecond
+	lowLatencyHeartbeatInterval  = 50 * time.Millisecond
+	defaultHeartbeatInitialDelay = time.Second
 )
 
 // HeartbeatTask is a perioic task to collect the heartbeat status from event dispatcher manager and push to heartbeatRequestQueue
@@ -42,7 +49,7 @@ func newHeartBeatTask(manager *DispatcherManager) *HeartBeatTask {
 		manager:    manager,
 		statusTick: 0,
 	}
-	t.taskHandle = taskScheduler.Submit(t, time.Now().Add(time.Second*1))
+	t.taskHandle = taskScheduler.Submit(t, time.Now().Add(heartbeatInitialDelay()))
 	return t
 }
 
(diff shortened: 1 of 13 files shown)


3AceShowHand (approved):

downstreamadapter/dispatchermanager/task.go
```diff
@@ -60,6 +66,20 @@ func (t *HeartBeatTask) Execute() time.Time {
 	return time.Now().Add(executeInterval)
 }
 
+func heartbeatInterval() time.Duration {
+	if config.GetGlobalServerConfig().IsLowLatencyMode() {
+		return lowLatencyHeartbeatInterval
```
wk989898, line 71: Can the HeartbeatInterval and HeartbeatInitialDelay parameters be set by the user?

logservice/eventstore/event_store.go
```diff
@@ -685,6 +685,9 @@ func (e *eventStore) RegisterDispatcher(
 
 	serverConfig := config.GetGlobalServerConfig()
 	resolvedTsAdvanceInterval := int64(serverConfig.KVClient.AdvanceIntervalInMs)
+	if serverConfig.IsLowLatencyMode() {
+		resolvedTsAdvanceInterval = 0
```
wk989898, line 689: ditto.

maintainer/maintainer.go
```diff
@@ -699,28 +701,33 @@ func (m *Maintainer) calCheckpointTs(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
```
wk989898, line 703: Should it be removed?

wk989898: And it looks like ticdc can't advance checkpointTs in no Low Latency Mode

PR #5903, opened 2026-08-07
debezium: output start_ts in the source block of DML messages

### LINKED ISSUE #5904: debezium: output the transaction start_ts in Debezium JSON messages
## Feature request

Expose the start TSO of the transaction that made a row change in the Debezium protocol output.

### Motivation

Debezium JSON messages currently expose only `commit_ts` in the `source` block (as a TiDB extension field). Downstream consumers that need to correlate a change event with its originating transaction (e.g. join with transaction-level metadata) have no way to recover the transaction's `start_ts` from the message, since the decoder approximates it with `commit_ts`.

### Proposal

- Emit `source.start_ts` (uint64 TSO, raw value) in the Debezium JSON value message, next to `commit_ts`, mirroring the existing behavior of `commit_ts` (always emitted).
- Declare `start_ts` in the source schema when `enable-tidb-extension` is enabled (JSON protocol only; Avro is out of scope).
- Make the Debezium decoder read `source.start_ts` and fall back to `commit_ts` when absent, preserving behavior for messages produced before this field existed.

### Acceptance criteria

- Insert/update/delete Debezium JSON messages carry `source.start_ts` equal to the event's `RowEvent.StartTs`.
- Decoder round-trip restores the true `StartTs`; old messages without the field still decode (fallback to `CommitTs`).
- Avro protocol and DDL/checkpoint messages are unchanged.

ekexium:

Issue Number: close #5904

Add a new changefeed-level option for Debezium JSON output (default `false`). Configure it with the sink URI parameter `debezium-include-start-ts=true` or in the changefeed config:

```toml
[sink.debezium]
include-start-ts = true
```

When enabled, the Debezium JSON value message carries the start TSO of the transaction that made the change as `source.start_ts`, next to `commit_ts`:

```
"source": {
  ...
  "commit_ts": 1,
  "start_ts": 5,      // new (only with debezium-include-start-ts=true)
  "cluster_id": "..."
}
```

- **EncodeValue**: write `source.start_ts = e.StartTs` when `DebeziumIncludeStartTs` is enabled.
- **writeSourceSchema**: declare `start_ts` under the same option (JSON only), so schema-validated consumers (e.g. Kafka Connect) can see the field **without `enable-tidb-extension`**. The JSON source-schema writer is shared with Avro, so the declaration is guarded against Avro — the Avro payload does not carry the field and its schema must not declare it.
- **decoder**: read `source.start_ts` and use it as the decoded event's `StartTs` (previously `StartTs = CommitTs`). Messages without the field (pre-feature format) fall back to `CommitTs`, preserving existing behavior. Present but invalid values (non-positive, out of `int64` range, or not an integer) emit an error log and fall back to `CommitTs` so consumption can continue.
- **config**: use `debezium-include-start-ts` in the sink URI, or `include-start-ts` under `[sink.debezium]` / the nested API v2 Debezium config. Explicit URI values (including `false`) override the config file — applied after the mergo merge because mergo cannot override a `*bool` `true` with `false`. Validation rejects the option for non-`debezium` protocols.

#### Tests 

- Unit test
- E2E tests

No measurable performance impact (one extra uint64 field per DML message, only when the option is enabled). With the default config (option off) nothing changes at all; with the option on, the `source` block gains one field and the schema declares it consistently, so schema-validated consumers see a matching declaration. DDL/checkpoint messages, key messages, row checksums and the Avro protocol are unchanged. The decoder falls back to `CommitTs` when `start_ts` is absent, so messages produced before this change decode exactly as before. Invalid present values are logged at error level and also fall back to keep production consumption running.

Yes — the TiCDC Debezium protocol and changefeed configuration documentation will be updated in a follow-up PR to describe both forms: the URI parameter `debezium-include-start-ts` and `[sink.debezium] include-start-ts`.

### Release note 

```release-note
Add an option to include the transaction start TSO in Debezium JSON messages. Configure it with the URI parameter `debezium-include-start-ts` or `[sink.debezium] include-start-ts`; when enabled, messages carry `source.start_ts`.
```

## Summary by CodeRabbit

- **New Features**
  - Added an option to include transaction start timestamps (`start_ts`) in Debezium JSON events.
  - Configure the option through sink settings or connection parameters.
  - Decoding supports `start_ts` and falls back to the commit timestamp when unavailable.

- **Improvements**
  - The option is validated for Debezium JSON only.
  - Avro, DDL, and checkpoint events continue to omit transaction start timestamps.


pkg/sink/codec/common/config_test.go
```diff
@@ -158,3 +158,29 @@ func TestDebeziumAvroWatermarkConfig(t *testing.T) {
 	require.True(t, cfg.AvroEnableWatermark)
 	require.Equal(t, "http://127.0.0.1:8081", cfg.AvroConfluentSchemaRegistry)
 }
+
+func TestDebeziumOutputStartTsConfig(t *testing.T) {
+	// URI parameter
+	cfg := NewConfig(config.ProtocolDebezium)
+	sinkURI, err := url.Parse("kafka://127.0.0.1:9092/topic?protocol=debezium&debezium-output-start-ts=true")
```
ginkgoch: Can we rename to `debezium-include-start-ts`? I remember we have `avro-include-before-value`, so I think to keep it consistent. https://github.com/pingcap/ticdc/pull/5154

ekexium: In the changefeed config, do you prefer `[sink] debezium-include-start-ts` or `[sink.debezium] include-start-ts = true`? #5154 is the in the former form, but I think the latter is more reasonable.

ginkgoch: Agree, please go with the latter one.

ekexium:
@wk989898 PTAL

pkg/sink/codec/debezium/codec.go
```diff
@@ -886,7 +886,10 @@ func (c *dbzCodec) writeBinaryField(writer *util.JSONWriter, fieldName string, v
 	writer.WriteBase64StringField(fieldName, value)
 }
 
-func (c *dbzCodec) writeSourceSchema(writer *util.JSONWriter, schemaName string) {
+// includeStartTs should only be true for DML row events: DDL and checkpoint
+// (watermark) messages have no per-row transaction, so their payloads never
+// carry start_ts and their schemas must not declare it.
+func (c *dbzCodec) writeSourceSchema(writer *util.JSONWriter, schemaName string, includeStartTs bool) {
```
wk989898, line 892: It looks like only dml will carry the start ts field?

ekexium: Yes, only DML events

pkg/sink/codec/debezium/codec.go
```diff
@@ -987,6 +990,16 @@ func (c *dbzCodec) writeSourceSchema(writer *util.JSONWriter, schemaName string)
 					writer.WriteStringField("field", "cluster_id")
 				})
 			}
+			// start_ts is gated by debezium-include-start-ts and declared only for
+			// the JSON protocol: the Avro payload does not carry it, so its schema
+			// must not declare it either.
+			if includeStartTs && c.config.DebeziumIncludeStartTs && !c.isDebeziumAvro() {
```
wk989898: I think "includeStartTs" is sufficient for judgment

pkg/sink/codec/debezium/codec.go
```diff
@@ -1176,7 +1191,7 @@ func (c *dbzCodec) EncodeValue(
 							jWriter.WriteRaw(fieldsJSON)
 						})
 					})
-					c.writeSourceSchema(jWriter, schemaName)
+					c.writeSourceSchema(jWriter, schemaName, c.config.DebeziumIncludeStartTs)
```
wk989898, line 1194: Does this affect the Debezium-Avro protocol?

ekexium: No. Debezium Avro explicitly calls `writeSourceSchema(writer, schemaName, false)` in `buildDebeziumAvroValueMessage`. And there is a test TestDebeziumConfluentAvroEncodeRowEvent

wk989898 (approved):

api/v2/model.go
```diff
@@ -1591,7 +1597,8 @@ type OpenProtocolConfig struct {
 
 // DebeziumConfig represents the configurations for debezium protocol encoding
 type DebeziumConfig struct {
-	OutputOldValue bool `json:"output_old_value" toml:"output-old-value"`
+	OutputOldValue bool  `json:"output_old_value" toml:"output-old-value"`
+	IncludeStartTs *bool `json:"include_start_ts,omitempty" toml:"include-start-ts,omitempty"`
```
lidezhu, line 1606: Preserve `output_old_value` when only `include_start_ts` is provided

A partial v2 update such as:

```
  {
    "replica_config": {
      "sink": {
        "debezium": {
          "include_start_ts": true
        }
      }
    }
  }
```

silently changes `output_old_value` from its default `true` to `false`. `UpdateChangefeed decode`s into an empty config, and `DebeziumConfig.OutputOldValue` is a non-pointer bool, so an omitted field becomes false. The conversion here then copies that zero value into a newly constructed internal `DebeziumConfig`, overriding the default. Consequently, Debezium update events stop carrying the before value even though the user only enabled start_ts.

Please make the API field optional (for example, `OutputOldValue *bool`) and only override the existing/default internal value when it is non-nil. Please also add a unit test to cover this case.

ekexium: Fixed, please take another look

lidezhu (approved):

PR #5907, opened 2026-08-07
eventstore: isolate tableStats by keyspace

### LINKED ISSUE #4998: [Bug] event_store tableStats is keyed only by TableID across keyspaces
## Description

In `logservice/eventstore/event_store.go`, `dispatcherMeta.tableStats` is indexed only by `TableID`:

- declaration: `tableStats map[int64]subscriptionStats`
- lookup: `e.dispatcherMeta.tableStats[dispatcherSpan.TableID]`
- insertion: `e.dispatcherMeta.tableStats[dispatcherSpan.TableID][subStat.subID] = subStat`

This assumes `TableID` is globally unique. However, that assumption does not hold across keyspaces: different keyspaces may have the same `TableID`.

As a result, subscriptions from different keyspaces can be placed into the same `tableStats` bucket.

## Current behavior

This does not appear to cause an incorrect match in the current code path, because the later span checks compare `StartKey` and `EndKey`, and those keys already include the keyspace prefix. The exact-match path also uses `TableSpan.Equal`, which checks `KeyspaceID` explicitly.

So this looks like a hidden pit / fragile invariant rather than an immediately reproducible bug.

## Why this is risky

The current implementation relies on an implicit rule: `tableStats` is keyed only by `TableID`, but correctness depends on all later matching logic remembering to compare the full span boundaries that carry keyspace information.

That makes the code easy to misuse in future refactors or new call paths, because the map key itself does not represent the full identity of a subscription span.

## Suggested improvement

Make `tableStats` keyspace-aware, for example by using a composite key such as `(KeyspaceID, TableID)`, or by using another comparable span identity that includes keyspace.

It would also help to add a unit test that registers the same `TableID` under different keyspaces and verifies they are not treated as candidates from the same logical bucket.

lidezhu:

Issue Number: close #4998 

- Isolate event-store subscription statistics by keyspace and table ID.
- Make TableSpan ordering keyspace-aware and distinguish complete span ranges.
- Filter reusable event-service candidates by keyspace.

#### Tests 

- Unit test
- Integration test
- No code

### Release note 

```release-note

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected span ordering to account for keyspace, table, start key, and end key.
  * Prevented subscriptions from different keyspaces from being conflated.
  * Ensured spans with different end keys are tracked and removed independently.

* **Tests**
  * Added coverage for keyspace-aware ordering, subscription isolation, and distinct span handling.

### DIFF
--- a/heartbeatpb/table_span.go
+++ b/heartbeatpb/table_span.go
@@ -17,13 +17,16 @@ import "bytes"
 
 // Less compares two Spans, defines the order between spans.
 func (s *TableSpan) Less(other *TableSpan) bool {
-	if s.TableID < other.TableID {
-		return true
+	if s.KeyspaceID != other.KeyspaceID {
+		return s.KeyspaceID < other.KeyspaceID
 	}
-	if bytes.Compare(s.StartKey, other.StartKey) < 0 {
-		return true
+	if s.TableID != other.TableID {
+		return s.TableID < other.TableID
 	}
-	return false
+	if cmp := bytes.Compare(s.StartKey, other.StartKey); cmp != 0 {
+		return cmp < 0
+	}
+	return bytes.Compare(s.EndKey, other.EndKey) < 0
 }
 
 func (s *TableSpan) Equal(other *TableSpan) bool {
--- a/heartbeatpb/table_span_test.go
+++ b/heartbeatpb/table_span_test.go
@@ -1,0 +1,53 @@
+// Copyright 2026 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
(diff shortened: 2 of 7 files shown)


lidezhu:
@coderabbitai review

3AceShowHand (approved):

heartbeatpb/table_span.go
```diff
@@ -17,13 +17,16 @@ import "bytes"
 
 // Less compares two Spans, defines the order between spans.
 func (s *TableSpan) Less(other *TableSpan) bool {
-	if s.TableID < other.TableID {
-		return true
+	if s.KeyspaceID != other.KeyspaceID {
+		return s.KeyspaceID < other.KeyspaceID
 	}
```
wk989898, line 22: Can `KeyspaceID` also be compared? What does this comparison mean?

lidezhu: Yes. `KeyspaceID` is a uint32, so it can be compared numerically. Here the comparison does not imply any semantic ordering between keyspaces. `Less` is used as the comparator for `BtreeMap`, and this establishes a deterministic lexicographical order on `(KeyspaceID, TableID, StartKey, EndKey)`. Including `KeyspaceID` ensures that otherwise identical spans from different keyspaces are treated as distinct B-tree keys.

wk989898 (approved):

PR #5956, opened 2026-08-12
redo: add DML two-stage ack

### LINKED ISSUE #3957: changefeed lag is more than 5s which is larger than disable redo log and larger than old-arch
### What did you do?

1、create changefeed and downstream is tidb
2、enable redo log
3、run sysbench

changefeed lag less than 5s

### What did you see instead?

changefeed lag is more than 5s which is larger than disable redo log

disable redo log

Image

enable redo log

Image

old arch with enable redo log

Image

### Versions of the cluster
./cdc version
 Release Version: v8.5.4-nextgen.202510.4-49-g20e0272
Git Commit Hash: 20e02729c294332deee3c258886d5ab33b436ecb
Git Branch: HEAD
UTC Build Time: 2026-01-05 20:36:58
Go Version: go version go1.25.5 linux/amd64
Failpoint Build: false
Kernel Type: Next Generation
2026-01-08T13:18:01.483+0800

### LINKED ISSUE #5936: redo: row-count flush threshold increases S3 checkpoint lag
### What did you do?

1. Run a TiCDC changefeed with redo logs stored in S3.
2. Upgrade TiCDC from v6.5.12 legacy architecture to v8.5.7 new architecture.
3. Run a recurring nightly workload containing large transactions and several
   `TRUNCATE` DDL statements.
4. Compare checkpoint lag and redo flush behavior before and after the upgrade.

Checkpoint lag should remain comparable to v6.5.12 under the same recurring
workload. Redo log files should normally be flushed according to
`max-log-size` or `flush-interval`, without excessive small-object requests to
S3.

DDL events following a large transaction should not be delayed significantly
by redo log persistence.

### What did you see instead?

After upgrading to v8.5.7 new architecture, checkpoint lag increases
significantly during the nightly workload. The problem did not occur with
v6.5.12 legacy architecture.

Observed behavior during the affected period:

- The new-architecture redo DML writer busy ratio remains at 100%.
- The redo flush duration sum rate remains around 0.95–0.98 seconds per second
  from 00:05 to 00:35.
- Each flush contains approximately 945–974 rows, close to the hard-coded
  1024-row threshold.
- Each redo file is only approximately 1.8–3.9 MiB, far below the configured
  64 MiB maximum log size.
- Average flush latency is approximately 142–251 ms.
- P90 flush latency is approximately 276–689 ms.
- P99 flush latency has a 1.5–4 second tail.
- DDL events are blocked behind preceding DML events in the ordered event path.

The new-architecture redo writer has a fixed row-count flush threshold near
1024 events. When this threshold is reached before `max-log-size` or
`flush-interval`, the writer synchronously waits for a small redo file to be
written to external storage.

For workloads with many relatively small row events, this serializes S3
request latency and produces many small objects. A large transaction can
require many such flushes before its callback completes, delaying the
checkpoint and subsequent barrier DDL events such as `TRUNCATE`.

### Versions of the cluster

```console
Not available in the current report.
```

```console
Not available in the current report.
```

```console
Before upgrade: v6.5.12, legacy architecture
After upgrade:  v8.5.7, new architecture
```

wk989898:

Issue Number: close #5936 close #3957

  Optimized the redo sink's two-stage acknowledgment pipeline:

  - Batches encoded redo events into multi-message spool entries, reducing per-row spool I/O and scheduling overhead.
  - Calls PostEnqueue only after the spool has accepted the entire batch, allowing the dispatcher to continue safely.
  - Keeps a separate PostFlush callback for each event and calls it only after the corresponding redo file has been persisted.
  - Releases each spool entry only after all events in that entry have been flushed.
  - Adds an ordered flush barrier when the spool disk quota is exhausted. Pending redo files are persisted and their quota is released before enqueueing more data.

  - Releases callbacks per completed redo file instead of waiting for every file in the flush round.

  This reduces spool overhead, prevents memory growth and quota-related stalls, and improves redo checkpoint latency and throughput.

#### Tests 

- Unit test
- Integration test
截屏2026-08-28 11 18 02
截屏2026-08-28 14 28 17

### Release note 

```release-note
 Improve redo sink throughput and checkpoint latency.
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable redo spool disk quotas and base directories with validation and isolated storage paths.
  * Added spool persistence, backpressure controls, quota handling, and monitoring metrics.
  * Added enqueue and flush progress tracking for redo events.
  * Added Grafana panels and labels for redo worker busy ratios.

* **Bug Fixes**
  * Improved callback timing, write coordination, quota reclamation, and writer cleanup.
  * Corrected metadata flushing to use the configured interval.

* **Refactor**
  * Consolidated redo writing and spool management into shared components.


3AceShowHand (approved):

pkg/redo/writer/encoding_worker.go
```diff
@@ -51,14 +53,15 @@ func toPolymorphicDMLEvent(
...
 	data := make([]byte, 8+len(rawData)+padBytes)
 	binary.LittleEndian.PutUint64(data[:8], lenField)
 	copy(data[8:], rawData)
 	return &polymorphicRedoEvent{
-		commitTs: rl.GetCommitTs(),
-		callback: event.PostFlush,
-		data:     data,
+		commitTs:    rl.GetCommitTs(),
+		postEnqueue: event.PostEnqueue,
+		postFlush:   event.PostFlush,
```
asddongmen: [P1] Avoid retaining raw row events through bound method callbacks

`event.PostEnqueue` and `event.PostFlush` are bound method values, so each closure retains the whole `*RedoRowEvent` receiver. After `ToRedoLog` has encoded the payload and PostEnqueue releases upstream quota, that receiver still keeps `RowChange` and its backing `chunk.Row` data alive until PostFlush. With slow or unavailable external storage, this lets raw rows accumulate in heap outside the spool memory accounting and defeats the intended 256 MiB hot-set bound.

I confirmed this with a `weak.Pointer` regression test: keeping only the returned polymorphic event kept the original `RedoRowEvent` reachable across repeated GCs. Please copy `event.EnqueueCallback` and `event.Callback` into the polymorphic event instead, with the required nil handling, so the encoded pipeline does not retain row payloads.

pkg/redo/writer/file_worker.go
```diff
@@ -384,12 +397,17 @@ func (f *fileWorkerGroup) flushAll(egCtx context.Context) error {
...
+	// Uploads still run concurrently in the flush workers.
 	for _, file := range f.files {
 		err := file.waitFlushed(egCtx)
 		if err != nil {
 			return errors.Trace(err)
 		}
+		for _, callback := range file.postFlush {
+			callback()
+		}
+		file.postFlush = nil
```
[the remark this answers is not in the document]
asddongmen: I checked this again at `e11ac3ff`, and it is still valid. A rotated file can finish `syncWriteFile` and close its `flushed` channel, but no goroutine consumes that completion until a later `flushAll`. If the current file remains partial, PostFlush and spool release therefore wait for the next ticker, quota barrier, or immediate flush even though the earlier file is already durable.

Please release the longest completed prefix from the background completion path while preserving file creation order. The current `TestFlushAllReleasesCallbacksPerCompletedFile` starts `flushAll` first, so it does not cover a rotated file that finishes before the next flush tick.

pkg/sink/spool/spool.go
```diff
@@ -620,17 +685,26 @@ func (s *Spool) Release(entry *Entry) {
 		return
 	}
 
+	releasedBytes := accountingBytes
 	if spilled {
+		releasedBytes = 0
```
lidezhu, line 713: If Truncate or Remove fails, `releasedBytes` remains zero, so the disk budget is not reduced and quota waiters are not awakened. Since the entry has already been cleared and there is no retry path for unreclaimed segments, a transient filesystem error can leave the changefeed blocked in WaitForDiskQuota indefinitely with only a warning logged.

lidezhu (approved):

pkg/sink/spool/spool.go
wk989898: Fixed. If Truncate or Remove fails, the segment now remains tracked and its disk quota is intentionally retained to match the actual disk usage. The spool schedules an asynchronous reclamation attempt immediately and retries every second while unreclaimed zero-reference segments remain.

ginkgoch (approved):

PR #6081, opened 2026-08-25
kafka: improve stability when creating many topics with Kafka v4

### LINKED ISSUE #6076: Kafka changefeed checkpoint lag exceeds threshold during concurrent workloads
## What did you do?

Ran a TiCDC next-generation Kafka/Avro changefeed with two concurrent upstream workloads.

Changefeed configuration:

- Kafka sink with Avro protocol and Schema Registry
- `partition-num=2`
- `replication-factor=2`
- `avro-enable-watermark=true`
- dispatcher topic template: `tidb_{schema}_{table}`
- integrity check level: `correctness`

The workloads were:

- `ycsb_bigtable`, one table, TPS 10000, 32 threads
- go-ycsb betting workload, 2000 tables, 32 threads

The test first captured an upstream TSO and confirmed that the changefeed checkpoint had advanced beyond it. It then continuously checked that checkpoint lag remained below 120 seconds, allowing at most 3 minutes of continuous violation.

All timestamps below are Asia/Shanghai (UTC+8).

The Kafka changefeed should remain healthy under the workload and keep checkpoint lag below the configured threshold.

If Kafka temporarily cannot serve a dynamically-dispatched topic or partition, the changefeed should either recover after the Kafka metadata/topic becomes available, or transition promptly to a clearly observable failed state with the sink error exposed to the test/operator.

## What did you see instead?

The changefeed repeatedly entered retry because Kafka message delivery failed.

- 2026-08-24 10:39:47: checkpoint-message delivery first failed with `kafka: Failed to deliver messages`.
- 2026-08-24 10:46:11: TiCDC reported `CDC:ErrKafkaAsyncSendMessage`; Kafka returned `Request was for a topic or partition that does not exist on this broker`.
- 2026-08-24 10:48:59: the test captured the upstream TSO.
- 2026-08-24 10:50:14: the checkpoint successfully advanced beyond that captured TSO.
- 2026-08-24 10:55:01 onward: Kafka unknown-topic/partition errors recurred. The changefeed checkpoint remained at TSO `468592865798783218` (physical time 10:52:03).
- 2026-08-24 10:58:49: another TiCDC node reported the same Kafka error. During dispatcher recovery, TiCDC logs showed a checkpoint/resolved-ts gap of about 6 minutes 40 seconds.
- 2026-08-24 10:59:33: the changefeed again entered backoff; retries were scheduled beyond the lag-check window.
- 2026-08-24 11:00:21: checkpoint-lag checking began. The lag stayed above 120 seconds for more than the 3-minute tolerance, causing the assertion to fail.

The test framework surfaced only a Ginkgo panic from the lag assertion, rather than the sampled lag value and the sink error:

```
step occurred panic, step: CDC checkpoint lag check,
threshold: 120, duration: 1h0m0s, toleration duration: 3m0s
```

Suspected: the Kafka sink error and retry behavior coincided with the checkpoint stall. The available artifacts do not establish why Kafka considered the topic or partition absent, so this issue does not attribute that underlying Kafka condition to TiCDC.

## Versions of the cluster

- TiCDC: `v26.3.3-9-g6500294`
- TiCDC commit: `65002945d670fea61492e35545326a0a289853c5`
- TiCDC build time: `2026-08-19 00:50:55 UTC`
- Architecture: TiCDC Next Generation
- Sink: Kafka with Avro protocol and Schema Registry

3AceShowHand:

Issue Number: close #6076

Kafka may acknowledge topic creation before every broker can return the new topic in metadata. This visibility window becomes easier to hit when a changefeed dynamically creates thousands of topics.

TiCDC previously suppressed `UnknownTopicOrPartition` from `GetTopicsMeta` even when the caller requested strict error handling. The topic manager could then read a missing metadata entry as zero partitions, cache the topic as ready, and start producing too early. The producer failure triggered repeated sink recovery and backoff, which could stall the changefeed checkpoint.

- Make `GetTopicsMeta` return topic-level errors, including `UnknownTopicOrPartition`, when `ignoreTopicError` is `false`. Discovery calls with `ignoreTopicError=true` continue to return metadata for valid topics.
- Retry temporary topic, broker, controller, and network metadata errors while waiting for a newly created topic to become visible. Authorization, authentication, configuration, and invalid-request errors fail immediately. Retries keep the existing six-attempt bound.
- Cache a topic and its partition count only after Kafka returns metadata for that topic.
- Mark the sink abnormal and close its input queues before releasing Kafka resources.
- Reject DML events after shutdown begins and honor cancellation before forwarding partitioned rows.

Together, these changes keep topic creation in the metadata-validation phase until Kafka confirms visibility and prevent the sink from accepting new DML events after shutdown begins.

#### Tests

- [x] Unit test
  - `GOTOOLCHAIN=go1.25.12 GOMAXPROCS=2 go test -p 1 -vet=off ./pkg/sink/kafka ./downstreamadapter/sink/topicmanager -count=1`
  - Covers topic metadata error handling, retry classification, and cache timing.
  - This branch also adds unit tests for sink shutdown ordering and post-close event handling.

Newly created topics may wait for up to the existing six metadata attempts during Kafka's visibility window; deterministic errors still return after the first attempt.

### Release note

```release-note
Fix Kafka changefeeds that can stall during large-scale dynamic topic creation when newly created topics are temporarily absent from broker metadata.
```


lidezhu (approved):

pkg/sink/kafka/admin.go
```diff
@@ -128,9 +128,6 @@ func (a *saramaAdminClient) GetTopicsMeta(topics []string, ignoreTopicError bool
 
 	for _, meta := range metaList {
 		if meta.Err != sarama.ErrNoError {
-			if meta.Err == sarama.ErrUnknownTopicOrPartition {
```
wk989898, line 131: Why remove this?

wk989898 (approved):

PR #6092, opened 2026-08-26
server: fence capture writes with etcd and P2P leases

### LINKED ISSUE #6045: TiCDC new arch: checkpoint advances past TSO but MySQL sink remains inconsistent after network loss
## What did you do?

Ran the `cdc_network_loss` integration test on a TiCDC new-architecture cluster
with a MySQL-compatible downstream sink.

Reproduction steps:

1. Create changefeed `cdc-network-loss-sync` to the downstream MySQL sink.
2. Pause the changefeed, prepare 100 sysbench tables with 500,000 rows each,
   then resume the changefeed.
3. Run `sysbench oltp_update_non_index` with 100 threads for one hour.
4. During a two-hour period, inject `NetworkChaos` every 30 minutes:
   - randomly select one upstream TiCDC peer;
   - packet loss: `100%`;
   - correlation: `25%`;
   - duration: `10s`.
5. After the workload and chaos period finish, get the current upstream TSO and
   wait until `ticdc_owner_checkpoint_ts` for the changefeed is greater than it.
6. Compare all upstream and downstream tables using CRC32.

Timeline (UTC):

- 16:10:17: resumed changefeed and started the one-hour workload.
- 16:40:17-27, 17:10:17-27, 17:40:17-27, 18:10:17-27:
  four 10-second network-loss injections.
- 17:10:18: workload finished.
- 18:10:28: captured the upstream TSO and started waiting for checkpoint.
- 18:10:58: checkpoint exceeded the target TSO; CRC comparison started.

Note: the current test creates NetworkChaos without an explicit `target`.
Because TiDB Cloud components use host networking, the framework warns that the
fault can affect the selected host's whole network scope rather than a single
TiCDC-to-peer link.

Internal execution: https://tcms.pingcap.net/dashboard/executions/plan/8226703

After network connectivity recovers, TiCDC should reconnect and replicate all
transactions up to the checkpoint. Once the changefeed checkpoint exceeds the
captured upstream TSO, CRC32 results between upstream and downstream should be
identical.

## What did you see instead?

The changefeed checkpoint exceeded the target TSO within 30 seconds, while the
workload had already stopped for one hour. However, CRC32 comparison remained
inconsistent for the full 90-minute retry window:

- 18:11:19: `workload.sbtest33`, source `1909784911`, sink `4228885997`
- 18:41:40: `workload.sbtest90`, source `3273849050`, sink `1173175951`
- 19:12:01: `workload.sbtest90`, same mismatch as above
- 19:42:22: `workload.sbtest8`, source `3724601131`, sink `2572728669`

The checker reports only the first mismatched table in each run; the table name
can vary because the comparison iterates over a Go map and stops at the first
mismatch. Therefore, the changing table name does not show that earlier tables
recovered.

The changefeed remained in normal state in the available logs. This indicates
that TiCDC's reported checkpoint can advance beyond the target TSO even though
the downstream data is not fully consistent.

## Versions of the cluster

- TiCDC: `v26.3.3-9-g6500294`
- TiCDC git hash: `65002945d670fea61492e35545326a0a289853c5`
- TiCDC build time: `2026-08-19 00:50:55 UTC`
- TiCDC kernel type: `Next Generation`
- TiDBX test plan: `v26.3.99-tidbx-release-20260820-hpmwh`
- Downstream: MySQL-compatible sink provisioned by the TiDBX testbed

asddongmen:

Issue Number: close #6045

A network-isolated capture may write after replacement dispatchers start. This overlap can leave a MySQL sink inconsistent after checkpoint advancement.

Gate capture writes with etcd and coordinator P2P proofs, negotiated through node heartbeats and a remote witness. Enforce admission in transport workers and before DML is sent to downstream. Confirmed session loss exits the capture. Add metrics, deterministic tests, and a 10-second removal grace.

Already-admitted transactions remain outside the guarantee because an accepted COMMIT cannot be revoked.

### End-to-end validation

A three-capture/MySQL test-infra gate sustained 20.3k DML/s, exercised lifecycle/network faults, matched ten table CRCs, and found no panic.

| Dimension | Previous `cdc_network_loss` | New synthetic chaos |
|---|---|---|
| DML | 100 threads; UPDATE only; no rate target | 256 threads; ~20k DML/s; UPDATE/DELETE/INSERT = 50%/25%/25% |
| Stress | 1h traffic; 10s fault every 30m | 2h15m traffic; 10s fault every 2m (15x frequency) |
| Coverage | Random capture, network loss only | 3 captures; ingress, egress, bidirectional, PD-only, PD-plus-CDC; scale 3→1→3, hang, kill |
| Validation | Checksum | CRC plus row-level Sync Diff for failed tables |

A six-round soak is running: 12 chaos hours, about 360 faults, and up to 972 million DMLs.

#### Tests

- Unit test
- Integration test
- Manual E2E test

The check is in memory; 500 ms heartbeats and capability negotiation preserve rolling upgrades.

No user documentation change; metrics expose write-lease state.

### Release note

```release-note
Fix a bug that may cause data inconsistent while there is a cdc node is network partitioned with other nodes. 
```

## Summary by CodeRabbit

* **New Features**
  * Added capture-wide write-lease protection for downstream writes.
  * Added P2P witness validation and etcd lease verification.
  * Applied write protection across MySQL, Kafka, Pulsar, cloud storage, and Redo outputs.
  * Added write-lease monitoring metrics and legacy-protocol compatibility.

* **Bug Fixes**
  * Prevented stale captures from publishing data, checkpoints, metadata, or cleanup changes.
  * Improved recovery after temporary lease loss and rejected stale or duplicate lease responses.

* **Documentation**
  * Added design documentation covering write-lease behavior, safety guarantees, and failure handling.


docs/design/capture-write-lease-design.md
```diff
@@ -0,0 +1,790 @@
...
+
+P2P improves stop-write latency during isolation, but it is not the replacement
+scheduling barrier.
+
+## 9. Failure behavior
+
+| Failure | Gate behavior | Capture exits? | Replacement behavior |
+| --- | --- | --- | --- |
+| Coordinator response is lost | P2P expires within five seconds and new writes stop. | No. | Capture key remains; no replacement is triggered. |
+| Coordinator capture is isolated from other nodes | Witness cannot ACK; local P2P expires. | Only if the etcd session is also confirmed lost. | Depends on capture-key deletion. |
```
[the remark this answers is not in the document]
asddongmen: Reviewed. The P2P subsection describes dual-proof admission, while the failure table already has a separate mixed-version row stating that P2P is not required and etcd proof remains required. No code change is needed.

coordinator/controller.go
```diff
@@ -527,6 +543,78 @@ func (c *Controller) onNodeChanged(ctx context.Context) {
...
+			hasGrant = true
+			break
+		}
+	}
+	delayed := false
+	failpoint.Inject("DelayCaptureWriteLeaseResponse", func(value failpoint.Value) {
+		if value.(bool) && hasGrant {
+			content, readErr := os.ReadFile(delayWriteLeaseResponseMarker)
+			delayMillis, parseErr := strconv.Atoi(string(content))
+			if readErr == nil && parseErr == nil && delayMillis > 0 &&
```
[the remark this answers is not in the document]
asddongmen: Resolved by commit 411f58b72. The marker files and strconv parsing were removed; the integration test now enables typed failpoints, and DelayCaptureWriteLeaseResponse consumes an integer value directly.

downstreamadapter/sink/kafka/sink_test.go
```diff
@@ -230,6 +231,59 @@ func TestKafkaSinkBasicFunctionality(t *testing.T) {
...
+	go func() {
+		runDone <- kafkaSink.Run(ctx)
+	}()
+	kafkaSink.AddDMLEvent(dmlEvent)
+
+	select {
+	case <-sent:
+		t.Fatal("Kafka DML was sent while the capture write gate was closed")
+	case <-time.After(100 * time.Millisecond):
+	}
```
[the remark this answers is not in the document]
asddongmen: Reviewed as a test-quality suggestion. These transport tests retain simple behavioral coverage, while deterministic gate transition and waiter behavior are covered in pkg/writelease. We are not adding production test hooks for these two integration-style unit tests.

maintainer/maintainer_manager.go
```diff
@@ -184,6 +195,13 @@ func (m *Manager) onCoordinatorBootstrapRequest(msg *messaging.TargetMessage) {
 			zap.Int64("version", req.Version))
 		return
 	}
+	if m.coordinatorID != msg.From || m.coordinatorVersion != req.Version {
+		m.writeGate.InvalidateP2P()
+		m.node.resetWriteLeaseRequests()
+	}
+	m.writeGate.SetP2PRequired(
+		req.GetWriteLeaseProtocolVersion() >= heartbeatpb.CurrentWriteLeaseProtocolVersion,
+	)
```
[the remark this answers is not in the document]
asddongmen: Fixed in commit 7d874868a. Bootstrap P2P activation now requires an exact protocol-version match, and TestCoordinatorBootstrapNegotiatesP2PWriteLease covers CurrentWriteLeaseProtocolVersion plus one falling back to etcd-only admission.

docs/design/capture-write-lease-design.md
```diff
@@ -0,0 +1,790 @@
...
+The server creates one capture-wide gate and publishes it through app context
+to dispatcher managers, ordinary sinks, and redo sinks. `Sink.SetWriteGate` is
+a mandatory interface method for every sink.
+
+Admission is checked in two stages:
+
+1. The outer `writeGatedSink` waits before an event enters the sink. This
+   provides backpressure and reduces unnecessary queue growth.
+2. Each transport checks again immediately before the operation that creates
+   the real downstream side effect. This closes the asynchronous-queue gap.
```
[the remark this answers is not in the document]
asddongmen: Reviewed. The final gate check is the admission point, and work that has already passed it is treated as in-flight. This PR does not promise cancellation of already-admitted work, so we are not adding a cross-transport handshake in this change.

tests/integration_tests/capture_write_lease/run.sh
```diff
@@ -0,0 +1,389 @@
...
+	# check tables are created and data is synchronized
+	for i in $(seq $DB_COUNT); do
+		check_table_exists "capture_write_lease_$i.usertable" ${DOWN_TIDB_HOST} ${DOWN_TIDB_PORT}
+	done
+	check_sync_diff $WORK_DIR $CUR/conf/diff_config.toml
+
+	# add more data in upstream and check again
+	for i in $(seq $DB_COUNT); do
+		db="capture_write_lease_$i"
+		go-ycsb load mysql -P $CUR/conf/workload2 -p mysql.host=${UP_TIDB_HOST} -p mysql.port=${UP_TIDB_PORT} -p mysql.user=root -p mysql.db=$db
```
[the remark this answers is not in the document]
asddongmen: No change is needed for the pinned go-ycsb v1.0.0 MySQL adapter. Both Insert and BatchInsert use INSERT IGNORE, so the overlapping keys are successful no-ops and the second load continues to the lease checks.

downstreamadapter/sink/mysql/sink.go
```diff
@@ -538,6 +550,7 @@ func (s *Sink) CleanupRemovedChangefeed() error {
 
 	cleanupWriter := mysql.NewWriter(context.Background(), -1, db, s.cfg, s.changefeedID, nil, nil)
 	defer cleanupWriter.Close()
+	cleanupWriter.SetWriteGate(s.writeGate)
```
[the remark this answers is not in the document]
asddongmen: Reviewed. Remove-only cleanup runs asynchronously and cannot block DispatcherManager TryClose or changefeed removal. A background context is intentional because the manager context is canceled before late cleanup runs; irreversible fencing also proceeds to process exit. No change is made here.

downstreamadapter/sink/pulsar/sink.go
```diff
@@ -316,6 +328,9 @@ func (s *sink) sendCheckpoint(ctx context.Context) error {
 				continue
 			}
 			common.SetCheckpointMessageLogInfo(msg, ts)
+			if !writelease.CanWrite(s.writeGate) {
+				continue
+			}
```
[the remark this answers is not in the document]
asddongmen: No change is needed for Pulsar. Its TopicManager.GetPartitionNum implementation always returns 1 and does not access Pulsar or create a topic. The existing later gate check remains immediately before actual message publication.

tests/integration_tests/synced_status_with_redo/run.sh
```diff
@@ -145,13 +145,11 @@ function run_normal_case_and_unavailable_pd() {
...
+	# case 2: when PD is unavailable, the capture loses its etcd write proof and
+	# exits instead of continuing to serve a potentially stale synced response.
+	cdc_pid=$(get_cdc_pid "$CDC_HOST" "$CDC_PORT")
 	kill_pd
-
-	sleep 20
-
-	synced_status=$(curl -X GET http://127.0.0.1:8300/api/v2/changefeeds/test-1/synced?keyspace=$KEYSPACE_NAME)
-	error_code=$(echo $synced_status | jq -r '.error_code')
+	ensure 30 "! kill -0 $cdc_pid > /dev/null 2>&1"
```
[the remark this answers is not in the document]
asddongmen: Fixed in commit 7d874868a. The integration case now rejects an empty or null CDC PID before killing PD, so the process-exit assertion cannot pass without a valid observed PID.

lidezhu (approved):

coordinator/capture_write_lease.go
```diff
@@ -0,0 +1,268 @@
...
+		coordinatorVersion: version,
+		selfNodeID:         selfNodeID,
+		now:                time.Now,
+		nonce:              rand.Read,
+		nodes:              make(map[node.ID]*captureLeaseNodeState),
+		p2pCapableNodes:    make(map[node.ID]struct{}),
+	}
+}
+
+func (c *captureWriteLeaseController) observeNodeCapability(id node.ID, version uint32) {
```
wk989898, line 74: Are these functions executed sequentially? Will it cause a data race?

coordinator/controller.go
```diff
@@ -511,11 +516,17 @@ func (c *Controller) onNodeChanged(ctx context.Context) {
 		zap.Any("removedNodes", removedNodes))
 
 	for _, n := range removedNodes {
+		if c.writeLease != nil {
```
wk989898: When will writeLease equal nil?

downstreamadapter/dispatchermanager/dispatcher_manager.go
```diff
@@ -145,13 +146,20 @@ type DispatcherManager struct {
...
 
 	// redo related
 	// redoEnabled is immutable and set to true if enabled.
 	redoEnabled bool
 	// redoReady set to true after the redo components are fully initialized and safe for concurrent access.
 	redoReady atomic.Bool
 	redoSink  *redo.Sink
+	// redoWriteSink is the capture-write-gated view used by redo dispatchers.
+	// redoSink remains the concrete sink used for lifecycle and redo metadata.
+	redoWriteSink sink.Sink
```
wk989898: I think it is not necessary to import the writeSink, because only one sink will be used. The wrapped sink is enough.

downstreamadapter/dispatcher/redo_dispatcher.go
```diff
@@ -98,6 +100,9 @@ func (rd *RedoDispatcher) SetRedoMeta(ctx context.Context, cfg *config.Consisten
 	}
 	ctx, rd.cancel = context.WithCancel(ctx)
 	rd.redoMeta = redo.NewRedoMeta(rd.sharedInfo.changefeedID, rd.startTs, cfg)
+	if gate, ok := appcontext.TryGetService[*writelease.Gate](appcontext.CaptureWriteGate); ok {
```
wk989898, line 103: I think it is unnecessary because the redo meta only updates after receiving a message from the maintainer

tests/integration_tests/capture_write_lease/run.sh
```diff
@@ -0,0 +1,437 @@
...
+
+function cleanup() {
+	disable_lease_failpoint_best_effort "$LEASE_DELAY_FAILPOINT"
+	disable_lease_failpoint_best_effort "$LEASE_DROP_FAILPOINT"
+	disable_lease_failpoint_best_effort "$LEASE_DUPLICATE_FAILPOINT"
+	stop_test "$WORK_DIR"
+}
+
+trap 'cleanup' EXIT
+run $*
```
wk989898, line 552: Please add a redo test

coordinator/controller.go
asddongmen: It should not be in normal situation. It's just a defender here.

coordinator/capture_write_lease.go
asddongmen: Yes. After construction, this controller is exclusively owned by the coordinator's single `runHandleEvent` goroutine. Message handlers only enqueue into `eventCh`, and the node-change callback only sets `nodeChanged.changed` under its own lock; all lease-state mutations run synchronously from `Controller.HandleEvent`. I documented this ownership invariant in fddb99e29. Focused race tests and the integration race-log check both pass.

coordinator/controller.go
asddongmen: You are right that it cannot be nil in production: `NewController` always initializes it and it is never cleared. The nil case only came from partial unit-test fixtures. In fddb99e29 I initialized the fixture and removed the runtime `writeLease == nil` / `writeLease != nil` branches so the constructor invariant is explicit.

downstreamadapter/dispatchermanager/dispatcher_manager.go
asddongmen: Agreed. There is only one underlying Redo sink, and DispatcherManager does not need any `*redo.Sink`-specific method. In fddb99e29, `redoSink` now stores the gated `sink.Sink` directly, and `redoWriteSink` plus `getRedoWriteSink` were removed. The regular sink still keeps its concrete/gated pair because MySQL recovery and cleanup require the concrete type.

downstreamadapter/dispatcher/redo_dispatcher.go
asddongmen: The maintainer fence and the write gate protect different moments. The maintainer message only updates in-memory unflushed metadata. `RedoMeta.Run` later performs external-storage writes and GC from independent timer goroutines, and initialization/cleanup also writes or deletes objects. A capture can receive a valid maintainer message, lose its write lease, and only then reach the timer-driven storage mutation. Therefore the gate still needs to remain at the actual external side-effect boundary. `TestRedoMetaFlushWaitsForWriteGate` covers this exact gap, and the new integration coverage verifies Redo progress stops and resumes with the capture gate.

tests/integration_tests/capture_write_lease/run.sh
asddongmen: Added in fddb99e29. The MySQL integration case now uses a small Redo-enabled changefeed for three probe tables. It waits until all capture gates reach `p2p_expired`, verifies persisted Redo resolved-ts stays below the post-write TSO while closed, and verifies it advances after lease recovery. It then deliberately blocks MySQL, stops CDC, applies Redo, and checks final consistency. On 10.2.15.7 the full case passed in 655s; Redo apply restored 300 rows, Sync Diff passed, and no data race was found.

tests/integration_tests/capture_write_lease/conf/write_lease_diff_config.toml
```diff
@@ -14,9 +14,9 @@ check-struct-only = false
     target-instance = "tidb0"
 
     target-check-tables = [
-        "capture_write_lease.lease_probe_1",
```
wk989898: Why remove the capture_write_lease database?

asddongmen: Thanks for asking. We did not remove the capture_write_lease database. It is still used by capture_write_lease.usertable and remains covered by sync-diff.

Only the three lease_probe tables were moved to capture_write_lease_redo. This database is exclusively owned by the new Redo-enabled changefeed, while the main changefeed explicitly excludes it. Keeping the two changefeed filters disjoint avoids duplicate replication and limits the Redo publication/apply checks to the small controlled probe workload instead of the long-running YCSB workload.

wk989898 (approved):

PR #6207, opened 2026-09-09
maintainer: avoid to create duplicate dispatcher when bootstrap

### LINKED ISSUE #6202: restored in-flight creates a duplicate dispatcher in maintainer_failover_when_operator
### What did you do?

Ran `tests/integration_tests/maintainer_failover_when_operator`

After failover the restored operators should converge: t6 / tableID 20 should end with exactly 1 replication after the restored merge completes, and no duplicate full-span dispatchers should run on the same table.

### What did you see instead?

 A newly created table (t4 / tableID 22) gets two full-span dispatchers on the same surviving node. One is cleaned up only after it starts writing, the cleanup races with a cloud-storage write and raises a retryable sink error, the whole changefeed is stopped/restarted, and the already-finished t6 merge is undone. After restart, t6 is re-added and auto-split into 2 spans (20 regions > region-threshold 10), so `wait_for_table_replication_count ... eq 1` times out.

### Timeline from logs

  - `12:35:20.223` (old maintainer epoch `468956868881678341`): t4 in-flight Create `154415...` is stored on the surviving node 8301.
  - `12:35:47.345` (new maintainer epoch `468956885016641547`): restore merge operator succeeds, and in the same ms:
    `maintainer_controller_bootstrap.go:384` logs
    `found a working table that is not in initial table map, just ignore it`,
    tableID=22.
  - `12:35:47.328` / `12:35:47.951`: node 8301 creates two dispatchers for the same t4 full span, tableID 22, identical start/end keys:
    - restored in-flight Create `154415...`
    - new Create `693762...` scheduled from the CREATE TABLE barrier under the
    new maintainer
- `12:35:49.599`: maintainer detects the extra dispatcher:
    `no span found, remove it`, dispatcherID=`154415...`.
  - `12:35:50.552`: while that dispatcher is still being removed/flushed, its
    cloudstorage writer fails:
    `CDC:ErrS3StorageAPI ... CDC_154415...json.tmp.<uuid>: no such file or directory`
  - `12:35:50.607`: maintainer treats it as `ErrChangefeedRetryable`, stops the
    changefeed, next retry `12:36:00.197`.
  - `12:35:49.755`: `merge dispatcher operator finished` for t6 — but the
  changefeed stop removes the merged dispatcher shortly after.
  - `12:36:00.557`: on restart, tableID 20 is re-added and immediately split:
    `split span by region count ... tableID: 20 ... spans=2 regionCount=20`.
    Table 20 keeps 2 replications and the `eq 1` check never passes.

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
master `b09792d0b1dd1385a21ea095bfa6ca288cd0c3d4` (next-gen)
```

wk989898:

Issue Number: close #6202 ref https://github.com/pingcap/ticdc/issues/5083

 Recover tables created after the bootstrap checkpoint from DDL history, preserving surviving dispatchers and split eligibility while repairing missing span coverage. This avoids duplicate scheduling during maintainer failover. 

#### Tests 

- Unit test
- Integration test

### Release note 

```release-note
 Fix an issue where maintainer failover could create duplicate table dispatchers and cause changefeed restarts.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bootstrap recovery for tables created or removed during initialization.
  * Restored cleanup operators and repaired task coverage for tables discovered during bootstrap.
  * Prevented adoption of tables dropped before the bootstrap checkpoint.
  * Added retry handling when table history is incomplete.

* **Tests**
  * Expanded coverage across default and redo bootstrap modes, operator restoration, task replacement, and paginated history handling.
  * Updated integration test selection for CI.

### DIFF
--- a/maintainer/maintainer_controller_bootstrap.go
+++ b/maintainer/maintainer_controller_bootstrap.go
@@ -148,7 +148,7 @@ func (c *Controller) FinishBootstrap(
 	}
 
 	// Step 4: Handle any remaining working tasks (likely dropped tables)
-	c.handleRemainingWorkingTasks(workingTaskMap, redoWorkingTaskMap)
+	c.handleRemainingWorkingTasks(allNodesResp, workingTaskMap, redoWorkingTaskMap)
 
 	// Step 5: Initialize route admission before barrier starts handling bootstrap
 	// block states. The barrier captures the route admin pointer at construction time.
@@ -373,8 +373,11 @@ func buildTableSplitMap(tables []commonEvent.Table) map[int64]bool {
 }
 
 func (c *Controller) handleRemainingWorkingTasks(
+	allNodesResp map[node.ID]*heartbeatpb.MaintainerBootstrapResponse,
 	workingTaskMap, redoWorkingTaskMap map[int64]utils.Map[*heartbeatpb.TableSpan, *replica.SpanReplication],
 ) {
+	c.adoptLeftoverWorkingSpans(redoWorkingTaskMap, common.RedoMode, collectReferencedBootstrapDispatcherIDs(allNodesResp, common.RedoMode))
+	c.adoptLeftoverWorkingSpans(workingTaskMap, common.DefaultMode, collectReferencedBootstrapDispatcherIDs(allNodesResp, common.DefaultMode))
 	for tableID := range redoWorkingTaskMap {
 		log.Warn("found a redo working table that is not in initial table map, just ignore it",
 			zap.Stringer("changefeed", c.changefeedID),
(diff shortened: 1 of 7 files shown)


maintainer/maintainer_controller_bootstrap.go
```diff
@@ -98,7 +98,17 @@ func (c *Controller) FinishBootstrap(
...
+	// Share the filter and schema store across snapshot and DDL history queries in both modes.
+	// Use an empty timezone because table filtering does not need it.
+	f, err := filter.NewFilter(c.replicaConfig.Filter, "", util.GetOrZero(c.replicaConfig.CaseSensitive), util.GetOrZero(c.replicaConfig.ForceReplicate))
+	if err != nil {
+		log.Error("load table from scheme store failed",
+			zap.String("changefeed", c.changefeedID.Name()),
+			zap.Error(err))
+		return nil, err
+	}
+	schemaStore := appcontext.GetService[schemastore.SchemaStore](appcontext.SchemaStore)
```
lidezhu, line 110: The maintainer should not access `SchemaStore` directly. These components communicate through the messaging layer, and introducing a direct reference breaks that architectural boundary and tightly couples their lifecycles.

maintainer/maintainer_controller_bootstrap.go
```diff
@@ -387,6 +434,83 @@ func (c *Controller) handleRemainingWorkingTasks(
...
+	for cursor := startTs; cursor < endTs && len(candidates) > 0; {
+		events, resolvedTs, err := schemaStore.FetchTableTriggerDDLEvents(c.keyspaceMeta, dispatcherID, f, cursor, 100)
+		if err != nil {
+			return nil, err
+		}
+		for _, event := range events {
+			for _, table := range event.NeedAddedTables {
+				if upperTs, ok := candidates[table.TableID]; ok && event.FinishedTs <= upperTs {
+					addedTables[table.TableID] = bootstrapAddedTable{table: table, startTs: event.FinishedTs}
+					delete(candidates, table.TableID)
```
asddongmen, line 499: I think we need to keep tracking the table until its candidate checkpoint instead of removing it after the first `NeedAddedTables`.

For example:

- the bootstrap snapshot is at ts=10 and table `t` does not exist;
- `t` is created at ts=20;
- `t` is dropped at ts=30;
- bootstrap reports the old dispatcher as `Stopped` with checkpoint=30.

At ts=20 this code adds `t` to `addedTables` and removes it from `candidates`, so the drop at ts=30 never invalidates it. Later `buildWorkingTaskMap` skips the stopped dispatcher, but `processTableSpans` creates a new absent span from ts=20. The scheduler may then recreate a dispatcher for a table that has already been dropped.

Could we replay both add and drop state through each candidate's upper timestamp, including normal and database-level drops, and only keep the tables that still exist there? A create-then-drop regression case would also help cover this.

PR #6254, opened 2026-09-14
 schemastore: fix replication when DDL makes a table eligible

### LINKED ISSUE #6243: After running ADD PRIMARY KEY to an ineligible table, TiCDC won't register a dispatcher for it, causing future DDLs depending on this table be blocked forever
### What did you do?

1. Create a cluster

    ```sh
    tiup playground-ng v8.5.8 --db 1 --pd 1 --kv 1 --tiflash 0 --ticdc 1 --ticdc.config ./ticdc-newarch.toml
    ```

2. Create a changefeed

    ```sh
    tiup cdc:v8.5.8 cli changefeed create -c test --sink-uri 'blackhole://'
    ```

3. Create an ineligible table

    ```sql
    CREATE TABLE test.a (pk BIGINT NOT NULL);
    ```

4. Add a primary key to this table

    ```sql
    ALTER TABLE test.a ADD PRIMARY KEY (pk);
    ```

5. Create another table that depends on it

    ```sql
    CREATE TABLE test.b LIKE test.a;
    ```

_No response_

### What did you see instead?

CheckpointTS is stuck after the last step.

### Versions of the cluster

v8.5.8

wk989898:

Issue Number: close #6243

  - Detect tables becoming eligible after DDL and notify the table trigger to create dispatchers, preventing subsequent CREATE TABLE LIKE from blocking.
  - Add DDL whitelist integration tests covering replication, filtering, and eligibility transitions.

**According to the [doc](https://docs.pingcap.com/tidb/stable/ticdc-ddl/#ddl-allow-list), there may be a risk of data inconsistency:**

When the upstream table has no valid index and force-replicate=true is not configured, the table will not be replicated. However, subsequent DDL statements (including CREATE INDEX, ADD INDEX, and ADD PRIMARY KEY) that create a valid index on this table will be replicated, which might cause inconsistency between downstream and upstream table schemas and lead to subsequent data replication failure.
#### Tests 

- Unit test
- Integration test
1. Create a cluster
```
tiup playground --db 1 --pd 1 --kv 1 --tiflash 0 --ticdc 1 --ticdc.binpath xxx --ticdc.config ./ticdc-newarch.toml
#Create a changefeed
tiup cdc:v8.5.8 cli changefeed create -c test --sink-uri 'blackhole://'
```
2. Create an ineligible table
```
CREATE TABLE test.a (pk BIGINT NOT NULL);
#Add a primary key to this table
ALTER TABLE test.a ADD PRIMARY KEY (pk);
#Create another table that depends on it
CREATE TABLE test.b LIKE test.a;
```
3. Changefeed lag is normal
截屏2026-09-14 19 12 41

### Release note 

```release-note
  Fix an issue where CREATE TABLE LIKE could stall replication after adding a primary or unique key to a previously ineligible table.
```

## Summary by CodeRabbit

- **New Features**
  - DDL replication now tracks both acquisition and loss of a table’s effective replication key.
  - Replication-key transitions are handled across default and forced replication modes, including partitioned tables and multiple schemas.
  - DDL events preserve replication-key state across storage reloads and remain compatible with previously persisted data.
  - Tables losing their final key are correctly removed from dispatching and downstream table scheduling.

- **Bug Fixes**
  - Improved historical schema lookup and handling of garbage-collected snapshots.
  - Preserved recovery behavior for replication-key loss during unfinished DDL operations.

- **Tests**
  - Expanded coverage for whitelist filtering, transitions, recovery, and dispatcher behavior.


logservice/schemastore/persist_storage_ddl_handlers.go
```diff
@@ -2323,6 +2325,19 @@ func buildDDLEventForDropTable(rawEvent *PersistedDDLEvent, tableFilter filter.F
 }
 
 func buildDDLEventForNormalDDLOnSingleTable(rawEvent *PersistedDDLEvent, tableFilter filter.Filter, tableID int64) (commonEvent.DDLEvent, bool, error) {
+	if rawEvent.TableBecameEligible {
+		// Without force replication, no dispatcher exists for the old schema.
+		// Let the table trigger execute the DDL and add the physical tables.
+		if tableFilter != nil && !tableFilter.IsForceReplicateEnabled() {
+			return buildDDLEventForNewTableDDL(rawEvent, tableFilter, tableID)
```
asddongmen, line 2392: [P1] Avoid emitting the re-eligibility DDL from both dispatchers

The assumption that no table dispatcher exists is not always true. For example, with `force-replicate = false`:

```sql
CREATE TABLE t (id BIGINT PRIMARY KEY NONCLUSTERED);
ALTER TABLE t DROP PRIMARY KEY;
ALTER TABLE t ADD PRIMARY KEY (id) NONCLUSTERED;
```

The `CREATE TABLE` creates a table dispatcher. Dropping the last effective key makes the table ineligible, but that DDL is filtered and the existing dispatcher is not removed. When the primary key is added back, `TableBecameEligible` is true, so this branch returns `buildDDLEventForNewTableDDL` to both the table trigger and the surviving table dispatcher. Both paths can then write the same DDL and report the same `NeedAddedTables`; MQ or cloud-storage sinks may emit duplicate DDL records.

Could we return the new-table event only when `tableID == common.DDLSpanTableID` in non-force mode, and skip it for ordinary table dispatchers? It would also be good to add a regression test for eligible -> ineligible -> eligible with an already registered dispatcher.

logservice/schemastore/persist_storage.go
```diff
@@ -390,6 +392,51 @@ func (p *persistentStorage) forceGetTableInfo(tableID int64, ts uint64) (*common
 	return store.getTableInfo(ts)
 }
 
+// getTableInfoForDDL reads the schema immediately preceding a DDL without
+// reconstructing every retained version for tables without local dispatchers.
+func (p *persistentStorage) getTableInfoForDDL(tableID int64, ts uint64) (*common.TableInfo, error) {
```
lidezhu: No need to create a new function and we can remove `forceGetTableInfo`. Also, `getTableInfoAtTs` should be a better name.

logservice/schemastore/persist_storage.go
```diff
@@ -759,6 +805,21 @@ func (p *persistentStorage) handleDDLJob(job *model.Job) error {
 		// ExtraTableInfo is the normal table info before exchange
 		ddlEvent.ExtraTableInfo, _ = p.forceGetTableInfo(ddlEvent.TableID, ddlEvent.FinishedTs)
 	}
+	switch job.Type {
```
lidezhu: It would be better to add an enrichPersistedDDLEventFunc hook that runs after releasing p.mu and before persisting the event, keeping historical table-info lookups inside the DDL handler abstraction without causing lock re-entry.
```
type persistStorageDDLHandler struct {
      ...
      enrichPersistedDDLEventFunc func(
          p *persistentStorage,
          event *PersistedDDLEvent,
      ) error
}

ddlEvent := handler.buildPersistedDDLEventFunc(...)
p.mu.Unlock()

if handler.enrichPersistedDDLEventFunc != nil {
      if err := handler.enrichPersistedDDLEventFunc(p, &ddlEvent); err != nil {
          return err
      }
}

writePersistedDDLEvent(p.db, &ddlEvent)
```

The eligibility-changing DDLs can share one enrichment function, and ExchangeTablePartition should also be moved into this hook with its lookup error propagated.

logservice/schemastore/persist_storage_ddl_handlers.go
```diff
@@ -1263,6 +1262,9 @@ func updateDDLHistoryForAddDropTable(args updateDDLHistoryFuncArgs) []uint64 {
 }
 
 func updateDDLHistoryForNormalDDLOnSingleTable(args updateDDLHistoryFuncArgs) []uint64 {
+	if args.ddlEvent.TableBecameEligible {
```
lidezhu: `TableBecameEligible` is ambiguous because eligibility also depends on changefeed filters and force-replicate, while this field only indicates that the table acquired a usable primary key or non-null unique key. A more precise name such as `TableAcquiredReplicationKey `would better describe this changefeed-independent schema transition.

lidezhu:
Please also handle the reverse transition for `DROP PRIMARY KEY` and other related DDL types, removing existing dispatchers when the table loses its last usable replication key while keeping them for force-replicated changefeeds.

asddongmen (approved):

lidezhu (approved):

PR #6258, opened 2026-09-14
schemastore: support FLASHBACK DATABASE DDL

### LINKED ISSUE #6095: schemastore: panic after DROP, FLASHBACK, and DROP DATABASE
### What did you do?

Run the following DDL sequence on the upstream TiDB cluster:

```sql
CREATE DATABASE test_db;
DROP DATABASE test_db;
FLASHBACK DATABASE test_db;
DROP DATABASE test_db;
```

The issue was observed on a five-node next-generation TiCDC deployment. All CDC nodes consumed the same DDL sequence and crashed when processing the second `DROP DATABASE`.

TiCDC should not crash. SchemaStore should either maintain the recovered schema metadata for `ActionRecoverSchema` or safely handle the subsequent `ActionDropSchema` while preserving correct DDL history.

### What did you see instead?

All CDC nodes panicked with the same stack:

```console
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10]

github.com/pingcap/ticdc/logservice/schemastore.updateDDLHistoryForSchemaDDL
github.com/pingcap/ticdc/logservice/schemastore/persist_storage_ddl_handlers.go:1068
github.com/pingcap/ticdc/logservice/schemastore.(*persistentStorage).handleDDLJob
github.com/pingcap/ticdc/logservice/schemastore/persist_storage.go:743
github.com/pingcap/ticdc/logservice/schemastore.(*keyspaceSchemaStore).tryUpdateResolvedTs
github.com/pingcap/ticdc/logservice/schemastore/schema_store.go:155
```

The crash repeats after each process restart because SchemaStore replays the same DDL sequence and encounters the same panic.

### Root cause analysis

The first `DROP DATABASE` removes the schema from `databaseMap`. `ActionRecoverSchema`, generated by `FLASHBACK DATABASE`, is explicitly ignored by `shouldSkipDDL`, so the schema metadata is not restored. The second `DROP DATABASE` then calls `updateDDLHistoryForSchemaDDL`, which dereferences the missing entry without checking it:

```go
for tableID := range args.databaseMap[args.ddlEvent.SchemaID].Tables {
```

This behavior is also present on the current `master` branch as of commit `3f0a68abf`.

Related tracking issue: #753 marks `ActionRecoverSchema` as ignored, but ignoring it is unsafe when later schema DDL depends on the recovered metadata.

### Suggested regression test

Add a SchemaStore test covering:

```text
CreateSchema -> DropSchema -> RecoverSchema -> DropSchema
```

The test should verify that SchemaStore does not panic, advances its resolved position, and retains consistent database/table metadata and DDL history.

### Versions of the cluster

Upstream TiDB cluster version:

```console
Not collected
```

Upstream TiKV version:

```console
Not collected
```

TiCDC version:

```console
Release version: v8.5.4-nextgen.202510.8
Git hash: 8ba50489f00e01097187acb0f071811593512267
```

The same unsafe code path was verified in `v26.3.1`, `v26.3.2`, `v26.3.3`, `v26.3.4`, `release-nextgen-202603`, `release-8.5`, and `master`.

lidezhu:

Issue Number: close #6095 

Add support for replicating FLASHBACK DATABASE DDLs, allowing TiCDC to restore and continue tracking a previously dropped database and its tables.

#### Tests 

- Unit test
- Integration test
- No code

### Release note 

```release-note

```

## Summary by CodeRabbit

- **New Features**
  - Added support for recovering entire databases, including tables, partitions, names, and metadata.
  - Recovered databases can be replicated through changefeeds using a dedicated “recover database” event.
  - Supports current and legacy database recovery formats.

- **Bug Fixes**
  - Corrected filtering for recovered databases and tables.
  - Restored table metadata is available immediately after schema recovery.
  - Improved execution of database recovery statements by MySQL sinks.

- **Tests**
  - Added coverage for recovery, filtering, replication, and legacy compatibility.


logservice/schemastore/multi_version.go
```diff
@@ -206,7 +206,7 @@ func (v *versionedTableInfoStore) doApplyDDL(event *PersistedDDLEvent) {
 	}
 	tableInfo, deleted := handler.extractTableInfoFunc(event, v.tableID)
 	if tableInfo != nil {
-		if ddlType == model.ActionRecoverTable {
+		if ddlType == model.ActionRecoverTable || ddlType == model.ActionRecoverSchema {
```
wk989898: I suggest using a switch case here

logservice/schemastore/persist_storage_ddl_handlers.go
```diff
@@ -101,8 +104,10 @@ type updateFullTableInfoFuncArgs struct {
 }
 
 type persistStorageDDLHandler struct {
+	// prepareJobFunc prepares job arguments before building the persisted DDL event.
+	prepareJobFunc func(storage *persistentStorage, job *model.Job) error
```
wk989898, line 108: Does only RecoverSchema need this function?

lidezhu: Yes.

logservice/schemastore/multi_version.go
lidezhu: Updated.

3AceShowHand (approved):

wk989898 (approved):

3AceShowHand (approved):

PR #6259, opened 2026-09-14
routing: fix rule validation, MQ dispatch and view rewriting

### LINKED ISSUE #6264: routing: incorrect rule matching and CREATE VIEW rewriting
### What did you do?

Used table routing in TiCDC's new architecture. The following configurations and DDLs expose independent problems:

- Set changefeed `case-sensitive = true` and route `Sales.*` while the source table is `sales.orders`.
- Create a changefeed with an invalid route matcher (`[`) and no source tables.
- Place a table-only route for `sales.orders` before an MQ rule for `sales.*` with topic `sales-events` and partition `ts`.
- Use `{schema}_{table}` with source schema `db{table}` and table `orders`.
- Route `test.*` to `dst.{table}_r` and create views using `SELECT t.* FROM t` or `WITH c AS (SELECT id FROM t) SELECT id FROM c`.

Routing should honor the changefeed configuration, validate matchers regardless of table count, preserve MQ dispatch behavior and literal source names, and preserve view query results.

### What did you see instead?

- The dispatcher uses the sink-level case-sensitivity field, producing different routes from API validation and the maintainer.
- Invalid matchers pass validation when the table set is empty.
- Table-only routing rules shadow MQ rules, changing the topic and partition selection.
- Placeholder substitution processes inserted source text again: `db{table}_orders` becomes `dborders_orders`.
- Qualified wildcards retain the old table name. CTE references are renamed as physical tables, such as `dst.c_r`, causing view creation failures or incorrect results when that physical table exists.

### Scope and likelihood

These issues require specific configuration combinations, unusual identifiers, or particular view definitions, so they are expected to be uncommon in typical production use. This assessment is based on the trigger conditions, not measured incident rates. When triggered, they can still cause incorrect routing or interrupt replication.

3AceShowHand:

Issue Number: close #6264

Also fixes stale table identities after EXCHANGE PARTITION, which can cause subsequent DML to use the previous table's route.

- Honor changefeed-level case sensitivity and validate route matchers even when no source tables exist.
- Skip table-only routing rules during MQ dispatch matching, preserving topic and partition selection by subsequent MQ rules.
- Substitute route placeholders in one pass so placeholder-like text in source names remains literal.
- Rewrite CREATE VIEW column and wildcard qualifiers with their physical tables, preserving aliases and CTE references across nested and recursive scopes.
- Build EXCHANGE PARTITION events with the current logical table identity and update participating dispatchers' cached table information so subsequent DML follows the current table's route.
- Emit the schema verification success log only after verification succeeds.

No configuration or event-format changes. Routing behavior changes for the affected cases described above. No performance benchmarks were run.

The dispatch-rule comments now clarify how table-only routing and matcher-only rules affect MQ dispatch. No new configuration or monitoring settings are introduced.

### Release note

```release-note
Fix table routing validation, case sensitivity, literal placeholder handling, and CREATE VIEW rewriting. Preserve MQ dispatch behavior for table-only routing rules and correctly route DML after EXCHANGE PARTITION.
```

## Summary by CodeRabbit

- **New Features**
  - Improved table routing for case-sensitive and case-insensitive identifiers.
  - Enhanced routing of views, correlated subqueries, CTEs, and database-level DDL.
  - Added support for continued DML routing after partition exchanges.
  - Added validation for empty routing targets and invalid route patterns.
  - Added table-state metadata for more reliable DDL event processing.

- **Bug Fixes**
  - Preserved changefeed bootstrap state during status updates and retries.
  - Missing changefeed deletion requests now require authentication.
  - Corrected handling of literal placeholder-like table names.
  - Improved dispatcher metadata handling during partition exchanges.


wk989898 (approved):

downstreamadapter/eventcollector/dispatcher_stat.go
```diff
@@ -493,7 +495,13 @@ func (d *dispatcherStat) updateTableInfoByDDL(ddl *commonEvent.DDLEvent) {
 		expectedTableID = current.(*common.TableInfo).TableName.TableID
 	}
 	if ddl.TableInfo.TableName.TableID != expectedTableID {
```
lidezhu, line 527: This fixes the immediate EXCHANGE PARTITION issue, but it leaks DDL-specific semantics into Event Collector. Event Collector should not need to know which DDL types can change the logical owner of a physical table, and BlockedTables should not be used to infer a schema mutation: it represents barrier participation, not whether the dispatcher's cached TableInfo must be replaced.

Could we model the per-dispatcher state transition explicitly in DDLEvent instead? For example:

```
type TableStateChangeKind uint8

  const (
      TableStateUnchanged TableStateChangeKind = iota
      TableStateUpdated
      TableStateDeleted
  )

  type TableStateChange struct {
      // The physical table ID whose dispatcher should apply this change.
      PhysicalTableID int64

      Kind  TableStateChangeKind
      After *common.TableInfo
  }

  type DDLEvent struct {
      // Existing event-level metadata used by DDL routing, rewriting, and sinks.
      TableInfo          *common.TableInfo
      MultipleTableInfos []*common.TableInfo

      // The state transition for the table dispatcher receiving this event.
      // It should be nil for the table-trigger event.
      TableStateChange *TableStateChange
  }

  Schema Store already has most of the required abstraction in extractTableInfoFunc:

  extractTableInfoFunc(
      event *PersistedDDLEvent,
      physicalTableID int64,
  ) (tableInfo *common.TableInfo, deleted bool)
```

Its return values already represent the three required states:

  tableInfo != nil              -> Updated
  tableInfo == nil && !deleted  -> Unchanged
  tableInfo == nil && deleted   -> Deleted

Instead of implementing the physical-to-logical table mapping again in individual buildDDLEventFunc implementations, fetchTableDDLEvents could attach the state change at a common boundary:

```
  handler := allDDLHandlers[model.ActionType(rawEvent.Type)]

  ddlEvent, ok, err := handler.buildDDLEventFunc(
      &rawEvent,
      tableFilter,
      tableID,
  )
  if err != nil || !ok {
      // existing handling
  }

  tableInfo, deleted :=
      handler.extractTableInfoFunc(&rawEvent, tableID)

  switch {
  case tableInfo != nil:
      ddlEvent.TableStateChange = &commonEvent.TableStateChange{
          PhysicalTableID: tableID,
          Kind:            commonEvent.TableStateUpdated,
          After:           tableInfo,
      }

  case deleted:
      ddlEvent.TableStateChange = &commonEvent.TableStateChange{
          PhysicalTableID: tableID,
          Kind:            commonEvent.TableStateDeleted,
      }

  default:
      ddlEvent.TableStateChange = &commonEvent.TableStateChange{
          PhysicalTableID: tableID,
          Kind:            commonEvent.TableStateUnchanged,
      }
  }
```

  This makes extractTableInfoFunc the single source of truth for both:

  1. updating versionedTableInfoStore; and
  2. describing the post-DDL state delivered to a table dispatcher.

The table-trigger event should continue carrying the event-level TableInfo and MultipleTableInfos required by routing and sinks, but it should not carry a per-table TableStateChange.

The router also needs to apply table routing to TableStateChange.After, just as it currently routes TableInfo and MultipleTableInfos, because Event Collector must cache the routed table identity:

```
  if ddl.TableStateChange != nil &&
      ddl.TableStateChange.After != nil {
      ddl.TableStateChange.After, err =
          router.ApplyToTableInfo(ddl.TableStateChange.After)
  }
```

With this contract, Event Collector becomes DDL-agnostic:

```
  d.tableInfoVersion.Store(ddl.FinishedTs)

  change := ddl.TableStateChange
  if change == nil {
      return
  }

  if change.PhysicalTableID != tableSpan.TableID {
      log.Error(
          "table state change was delivered to the wrong dispatcher",
          zap.Int64("expectedPhysicalTableID", tableSpan.TableID),
          zap.Int64("actualPhysicalTableID", change.PhysicalTableID),
      )
      return
  }

  switch change.Kind {
  case commonEvent.TableStateUnchanged:
      return

  case commonEvent.TableStateUpdated:
      if change.After == nil {
          log.Error("updated table state does not contain TableInfo")
          return
      }
      d.tableInfo.Store(change.After)

  case commonEvent.TableStateDeleted:
      // Dispatcher removal is handled by the scheduling/barrier workflow.
      return
  }
```

  Please also include TableStateChange in:

  - DDLEvent marshal/unmarshal;
  - routed-event cloning;
  - size accounting, if applicable;
  - mixed-version compatibility handling.

  The tests should verify the contract at the Schema Store boundary:

  ALTER TABLE                         -> Updated
  CREATE VIEW broadcast              -> Unchanged
  CREATE TABLE LIKE referenced table -> Unchanged
  EXCHANGE: old normal physical ID   -> Updated to partition table
  EXCHANGE: old partition ID         -> Updated to normal table
  DROP TABLE                          -> Deleted

lidezhu (approved):

nongfushanquan (approved):

PR #6265, opened 2026-09-15
api,cli: support creating paused changefeeds

### LINKED ISSUE #6266: Support creating changefeeds in a paused state

Creating a changefeed currently starts replication immediately. Some workflows need to validate and save the changefeed and protect its starting position, then explicitly start replication later.

Support an optional `pause` parameter on `POST /api/v2/changefeeds` and top-level `pause = true` in the CLI TOML configuration.

- Omitted or `false`: preserve the existing automatic startup behavior.
- `true`: perform the normal configuration, table, sink, TSO and GC safety checks, then persist the changefeed in the existing `stopped` state without scheduling replication.
- Initialize the checkpoint to the validated start-ts and retain normal paused-changefeed GC protection and TTL limits.
- Keep the changefeed stopped across restarts. The existing resume API/command starts replication from its saved checkpoint unless the caller explicitly overrides it.
- Treat this as a create-only option; configuration updates must not implicitly pause or resume an existing task.

Creating a running changefeed and immediately pausing it leaves a window in which replication can already start. A paused initial state avoids that window.

API callers can add `"pause": true` to the creation request. CLI users can put `pause = true` before any section headers in their TOML file and later run `cdc cli changefeed resume --changefeed-id <id>`.

Existing callers do not need to change. This does not introduce a new lifecycle state or unlimited data retention.

Implementation PR: https://github.com/pingcap/ticdc/pull/6265

sdojjy:

Creating a changefeed immediately schedules replication. Callers need to perform the normal validation and reserve the start position while leaving the new task paused until an explicit resume.

Issue Number: close #6266

- Add the optional top-level `pause` boolean to the v2 create request. Omission and `false` retain automatic startup; `true` persists `stopped` and registers the task in the stopped collection without scheduling a maintainer.
- Support top-level `pause = true` in CLI TOML via the internal replica configuration and forward it to the create request. It is a create-only option, not an additional runtime state or nested API replica setting.
- Keep the existing table/sink/configuration/TSO checks, initial checkpoint, temporary GC protection and coordinator GC handoff. Persist an internal pending-bootstrap marker so the first resume and coordinator restart retain fresh-changefeed semantics, including stale MySQL DDL recovery record cleanup. Clear it only after the current maintainer reports successful bootstrap, using an incarnation/epoch check and a metadata revision comparison. Failed acknowledgements are retried on subsequent heartbeats; later resumes keep normal recovery semantics.
- Add creation/resume/GC unit coverage and a MySQL integration case covering API and CLI creation, default compatibility, rejected requests, no replication while paused, restart, GC handoff on classic, and data consistency after resume with newer stale `ddl_ts_v1` records for a reused name. Register the case in MySQL light CI G04 and document usage.

#### Tests

Passed locally:

- Full `./coordinator`, `./coordinator/changefeed`, and `./coordinator/operator` package tests with `-p 1 -ldflags='-s -w' -count=1`, including first resume/restart, stale owner, failed acknowledgement retry, and metadata compare-and-swap coverage.
- MySQL `TestGetTableRecoveryInfo_StartTsGreaterThanDDLTs` and `TestGetTableRecoveryInfo_RemoveDDLTs`; local targeted `unconvert` lint.

- `go test -p 1 -ldflags='-s -w' ./api/v2 ./cmd/cdc/cli ./pkg/config -count=1`
- `go test -p 1 -ldflags='-s -w' ./coordinator -run '^Test(CreateChangefeedDoesNotUpdateGCSafepoint|UpdateGCSafepointCallsGCManagerUpdate)$' -count=1`
- Repository-pinned Go/import/shell formatters on changed files, shell syntax checks, diff line-width and log-style checks, and `git diff --check`.

Not completed locally:

- The same Coordinator tests with `-tags=nextgen` could not compile because the disk ran out of space.
- `make integration_test_mysql CASE=changefeed_create_paused` stopped at `check_third_party_binary`: local database test binaries are missing, and the Docker daemon is unavailable. The new integration test is committed and registered in CI but has not been run end-to-end locally.
- Full binary build and full repository/race suites were not run given disk capacity. An initial broad `-run TestCreateChangefeed` also hit an existing test's missing global MessageCenter initialization; the focused GC creation tests above initialize their own services and pass.

#### Will it cause performance regression or break compatibility?

The default creation path is unchanged. No new lifecycle state, storage migration, or scheduler abstraction is introduced; existing metadata without the optional bootstrap marker retains its current behavior. Pause affects creation only, and existing paused-task GC TTL limits still apply.

#### Do you need to update user documentation, design documentation or monitoring documentation?

Added `docs/create-paused-changefeed.md` with API and CLI TOML examples and lifecycle/GC semantics.

### Release note

```release-note
Support creating paused changefeeds through the v2 API pause option and CLI TOML configuration while preserving validation and GC protection.
```

## Summary by CodeRabbit

* **New Features**
  * Changefeeds can now be created in a paused state through the REST API or CLI using the `pause` option.
  * Paused changefeeds remain stopped across restarts and can be resumed later while preserving fresh-start behavior.
  * Omitting `pause` or setting it to `false` retains automatic startup.
  * Existing validation and GC protections continue to apply.

* **Documentation**
  * Added guidance for configuring and using paused changefeed creation.

* **Tests**
  * Added coverage for API, CLI, restart, validation, resumption, and replication scenarios.


asddongmen:
I think we need to preserve the new-changefeed semantics for a changefeed created with `pause = true`.

The changefeed is constructed with `isNew = true`, but its first normal resume goes through `MoveToSchedulingQueue`, which sets `isNew` to `overwriteCheckpointTs` and therefore turns it into `false`. If the coordinator restarts before the first resume, bootstrap also reconstructs the stopped changefeed as not new.

This matters for MySQL sinks because `IsNewChangefeed` is eventually used as `removeDDLTs`. With the flag set to false, the sink reuses existing `ddl_ts_v1` recovery records instead of clearing them. If a changefeed ID is reused and then created paused with a historical `start_ts` earlier than stale `ddl_ts`, the effective start position can be moved forward and some events may be skipped.

Could we keep a durable marker for a paused changefeed that has never started, preserve it across coordinator restart, and clear it only after the first maintainer bootstrap succeeds? It would also be good to assert `IsNewChangefeed` on the first resume and cover a stale `ddl_ts` case.

wk989898 (approved):

asddongmen (approved):

nongfushanquan (approved):

PR #6284, opened 2026-09-18
config: add allow-same-cluster with a static routing validation

3AceShowHand:

Issue Number: close #6285

Support copying data into separate databases within the same TiDB logical cluster through an explicit `allow-same-cluster` option, while preventing replication from writing back into the source range.

Table-level separation alone is insufficient: with `a.t1 → b.t1_copy` and `b.t2 → c.t2_copy`, the target tables are outside the filter, but replicating `DROP DATABASE a` would delete source database `b`.

- Add the optional changefeed setting `allow-same-cluster` and its API v2 counterpart `allow_same_cluster`. The default same-cluster rejection remains unchanged.
- Validate the effective configuration before this option takes effect on create, update, and resume. Invalid configurations return `CDC:ErrInvalidReplicaConfig`.
- Require database isolation:
  1. Each complete `schema.table` filter rule must be covered by a routing matcher.
  2. Every matching route must send the source database outside the filter's database range, including database DDL and future databases.
  3. Routes overlapping on a source database must have identical `target-schema` expressions. Database DDL considers every matching schema rule, including rules with disjoint table matchers.
- Parse each matcher into an individual validation entry. Check target overlap per entry and schema consistency between entries. Database isolation makes separate table-target conflict searches unnecessary; the witness search immediately skips fixed targets outside the source range.
- Validate update requests on a cloned changefeed configuration so rejected requests cannot mutate coordinator state. The no-routing fixture explicitly clears dispatchers because CLI configuration updates merge with existing values.
- Match runtime handling of rule padding and case sensitivity, preserving whitespace inside quoted names and parsing placeholders before normalizing target literals.

Accepted example: `src.* → dst.*`, when `dst` is outside the source database range. Rejected examples: `src.t1 → src.t1_copy`, the `a → b → c` source chain above, and `*.orders → {schema}_backup.orders_copy`.

Supported rule parts are literals, `*`, a single leading/trailing wildcard, and supported quoted names. Targets support literal text around one corresponding `{schema}` or `{table}` placeholder. Unsupported syntax is rejected. This is a conservative contract: coverage must come from a single matcher, and different schema expressions are rejected on overlapping source ranges even if they happen to produce the same name.

Review entry points:

- `pkg/check/same_cluster.go`: parsing, coverage, database isolation, and schema-routing consistency.
- `pkg/check/same_cluster_runtime_test.go`: comparison against the real filter and DDL router.
- `tests/integration_tests/same_upstream_downstream/`: eight configurations covering normal CLI operations, rejected create/update requests, state checks, continued replication after rejected updates, and database CREATE/ALTER/DROP routing.
- `docs/design/2026-09-18-allow-same-cluster-gate-design.md`: examples, supported inputs, and conservative limitations.

#### Tests

- Unit tests: `go test -race -tags=intest ./pkg/check -count=1` passed; failpoint injection was not enabled.
- Configuration/API tests: `go test ./pkg/config ./api/v2 -run 'TestReplicaConfig_AllowSameCluster|TestReplicaConfigConversion' -count=1` passed.
- API/CLI regression tests passed with race detection: `go test -race -tags=intest ./api/v2 ./cmd/cdc/cli -run 'TestRejectedUpdatePreservesChangefeedInfo|TestSameClusterUpdateConfigs|TestApplyChanges|TestChangefeedUpdateCli' -count=1`. These verify rejected updates preserve stored configuration and all seven invalid fixtures produce their expected errors after CLI merging.
- Integration tests: updated `same_upstream_downstream`; not executed locally. They use configuration files and CLI operations, with no direct metadata edits. Each rejected update is followed by a configuration query to confirm the original configuration is unchanged.
- Checked all eight TOML configurations against the validator, exercised database DDL rewriting with the real router, and checked shell syntax and formatting.
- `make fmt` and `git diff --check` passed.
- Go benchmarks cover validation cost separately from integration tests.

The option is new and disabled by default. Opt-in requires cross-database isolation; same-database table renaming is unsupported. Validation operates on configuration patterns at create/update/resume time, without the previous schema/table candidate cross product. Multi-rule configurations still require overlap checks.

The design document describes the database-isolation contract and its limitations. User-facing documentation should explain the new option and supported routing configurations. No monitoring changes.

### Release note

```release-note
Add an opt-in allow-same-cluster option for replication into isolated databases within the same TiDB cluster.
```

## Summary by CodeRabbit

* **New Features**
  * Added the `allow-same-cluster` option for changefeeds, enabling same-cluster replication when safe table routing is configured.
  * Added API and configuration support, including case-sensitive and schema-routing scenarios.
* **Bug Fixes**
  * Invalid routing configurations are now rejected with clear validation errors.
  * Rejected changefeed updates no longer leave partial configuration changes behind.
* **Documentation**
  * Added guidance covering supported routing patterns and safety requirements.


ginkgoch (approved):

pkg/check/same_cluster.go
```diff
@@ -0,0 +1,506 @@
...
+	if err := verifyFilterRulesCovered(filters, routes); err != nil {
+		return err
+	}
+	if err := verifySchemaTargets(filters, routes); err != nil {
+		return err
+	}
+	return verifySchemaRoutingConsistent(filters, routes)
+}
+
+func normalizeName(name string, caseSensitive bool) string {
```
lidezhu, line 74: With `CaseSensitive=true`, a configuration such as `filter/matcher src.*` with target-schema = "SRC" passes this validation because the target is compared case-sensitively. TiDB only supports `lower_case_table_names=2`, so SRC resolves to the existing src database in the same logical cluster. Sink writes therefore land back in the source schema and become eligible for recapture, defeating the isolation guarantee and potentially causing repeated writes or changefeed failure. Please make schema-isolation comparisons account for TiDB's case-insensitive identifier lookup regardless of the TiCDC filter setting, or reject `allow-same-cluster` when case-sensitive filtering is enabled, and add an integration regression case that exercises this through TiDB rather than only the in-memory filter/router.

docs/design/2026-09-18-allow-same-cluster-gate-design.md
```diff
@@ -0,0 +1,82 @@
...
+
+`src.* → dst.*` 在 `dst` 不匹配源库范围时仍然允许。`isolation_src*.* → copy_{schema}.{table}` 也允许，未来新建 `isolation_src_future` 的数据库 DDL 会路由到 `copy_isolation_src_future`。
+
+数据库隔离保护 filter 定义的源范围。目标库仍会正常执行复制产生的数据库 DDL，用户需要为目标库安排相应的数据所有权。
+
+## 输入契约与保守限制
+
+支持的 filter/matcher 为 `schema.table`，每段支持字面量、`*`、单个前缀或后缀通配，以及支持范围内的引号名字。仅去掉规则外围空格和制表符，保留引号内空白。未配置 filter 规则按 `*.*` 处理。
+
+目标支持字面量或带一个对应占位符的表达式：`target-schema` 使用 `{schema}`，`target-table` 使用 `{table}`；空目标表示保留原名。虽然目标表不参与隔离推理，仍校验其表达式是否属于支持范围。大小写按 `case-sensitive` 配置处理，先解析占位符再规范化目标字面文本。
```
lidezhu: This case-sensitivity description is now stale. The latest implementation intentionally preserves the configured case-sensitive semantics for filter/matcher coverage, but checks target-schema isolation case-insensitively because TiDB identifier lookup uses lower_case_table_names=2. Please document these two distinct behaviors and the TiDB rationale so future changes do not reintroduce the case-only routing bypass.

lidezhu (approved):

wk989898 (approved):

PR #6314, opened 2026-09-20
kafka consumer: improve downstream write throughput

### LINKED ISSUE #6315: kafka consumer: improve downstream write throughput
### What did you do?

Run Kafka integration tests using the Kafka consumer to write decoded events to a MySQL-compatible downstream. The test setup forces `safe-mode=true&batch-dml-enable=false` on the consumer's downstream sink.

The consumer should use the MySQL sink defaults, including batched DML, to improve downstream write throughput. It should preserve row identity, column types, event ordering, and correctness when upstream events are replayed.

### What did you see instead?

The forced safe mode and disabled batched DML limit consumer write throughput. Removing these overrides exposes decoder and consumer assumptions that must be addressed for the default write path:

- **Row identity:** reconstructed metadata can contain incorrect column offsets or incomplete primary-key indexes. A composite primary key can be represented as `PKIsHandle`, causing UPDATE/DELETE predicates to use the wrong column. For example, a table with `PRIMARY KEY (s_i_id, s_w_id)` can be matched on `s_data` when that column sorts first, risking missed or incorrect row updates.
- **Column types and buffered events:** decoded row chunks must match the table metadata used by the writer, including ENUM/SET representation. Buffered events must retain their own schema when a later DDL changes the decoder's cached metadata.
- **Avro events:** deletes may only carry key fields, and updates may omit before values. The consumer still needs sufficient metadata and event semantics to write these changes correctly.
- **Upstream replay:** table scheduling can produce `A(10), B(20), A(10), B(20), watermark`. Passing the replayed rows directly to the batch writer can trigger its duplicate-insert panic. The consumer should emit `A(10), B(20)` while retaining distinct legitimate row changes. Events exactly at the watermark may still be replayed.

The composite-key finding is based on decoder metadata and SQL-generation analysis. No quantitative throughput improvement is claimed; the objective is to enable the existing faster write path safely.

### Versions of the cluster

- TiCDC: master; the original composite-key analysis used `d1a3a8dd1ee5297cb087835a4c0dadfde9de0080`.
- Upstream TiDB/TiKV versions are not required to reproduce the decoder-level metadata issue.

3AceShowHand:

Issue Number: close #6315

Kafka integration tests force `safe-mode=true&batch-dml-enable=false` on the consumer's downstream MySQL sink, limiting write throughput. This PR removes those overrides so the consumer can use the default unsafe mode with batched DML to write faster. The accompanying decoder and consumer changes adapt event metadata and replay handling to that write path.

- Remove `safe-mode=true` and `batch-dml-enable=false` overrides from Kafka consumer test setup and integration scripts, using the MySQL sink defaults.
- Reconstruct complete primary-key and handle-key metadata, including column offsets and index properties, so the default writers can identify rows correctly across supported decoders and storage schema files.
- Reuse Open Protocol table metadata within DDL intervals while preserving the schema and column types needed by buffered events. Reuse previously seen Avro value schemas for deletes, and represent updates without before values as keyed DELETE + INSERT operations.
- Deduplicate replayed rows after spill restoration and decoding, before submitting them to the sink. Match table, commit-ts, operation, and handle key; retain watermark-boundary state until it is safe to discard. This prevents replay from violating the batch writer's row-change assumptions.

#### Tests

- Targeted unit tests for consumer replay invariants and decoder metadata/type handling passed locally.
- Kafka integration CI: the light suite passed; the heavy suite is pending at the time of this update.
- CI is not fully green: the check and unit-test jobs currently report failures.
- Full local spill validation is blocked by the existing filesystem usage limit.

The intended performance improvement comes from using batched DML and avoiding forced safe mode in the test consumer. No throughput benchmark has been run, so no quantitative speedup is claimed. The supporting changes preserve row identity, schema correctness, and replay semantics for this faster write path. No wire-format changes or MySQL sink logic changes are introduced.

No.

### Release note

```release-note
None
```


pkg/common/table_info.go
```diff
@@ -502,6 +503,28 @@ func (ti *TableInfo) IsEligible(forceReplicate bool) bool {
 	return ti.HasPKOrNotNullUK
 }
 
+// SetHandleKeyFlags records how the primary key is stored as the row handle,
```
wk989898, line 506: Indicate this is for testing

wk989898 (approved):

asddongmen (approved):

